PR #10, opened 2017-02-16
Add .gitignore

stshine:
Add gitignore file to ease development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,11 @@
+/target
+Cargo.lock
+
+# Editors
+
+# IntelliJ
+.idea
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -0,0 +1,11 @@
+/target
+Cargo.lock
+
+# Editors
+
+# IntelliJ
+.idea
```
emilio: I believe these belong to a global gitignore file from the user though.

stshine: Well, this part comes from the gitignore file of servo, so I guess it should be OK?

PR #19, opened 2017-04-01
CI builds

jdub:
- Travis CI tests Rust stable, beta, and nightly on Linux and macOS
- Appveyor tests Rust stable and beta on Windows with MSVC and MSYS/GNU
- Badges for both

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,44 @@
+sudo: false
+language: rust
+cache: cargo
+
+os:
+  - linux
+  - osx
+
+rust:
+  - stable
+  - beta
+  - nightly
+
+addons:
+  apt:
+    sources:
+      - george-edison55-precise-backports
+    packages:
+      # Newer cmake required on precise
+      - cmake
+      - cmake-data
+      # Pathfinder requires some X libraries
+      - libxcursor-dev
+      - libxinerama-dev
+      - libxrandr-dev
+      - libxxf86vm-dev
+
+env:
+  global:
+    - TRAVIS_CARGO_NIGHTLY_FEATURE=""
+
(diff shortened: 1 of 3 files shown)


jdub:
During this month's Bay Area Rust Meetup (which I watched via Air Mozilla), you made an offhand comment about desiring CI support. Presto! The Internet delivers. ✨

**Notes:**

Neither configuration runs tests on stable, because you're using unstable test features. A quick tweak if and when you get around to it: Just drop `--only nightly`.

Both configurations build docs for stable, but don't do anything with them. It's a good thing to test, and with https://docs.rs/ we don't need to upload documentation anymore anyway.

The Appveyor configuration doesn't build nightly, and thus doesn't run tests, due to a [bug in travis-cargo on Windows](https://github.com/huonw/travis-cargo/issues/33). I suspect travis-cargo hasn't been used on Windows before, but I decided to use it in support of community infrastructure.

The Appveyor configuration produces a build artifact of the lorem-ipsum example. I'll probably follow up with a quick patch to include Nimbus in the binary, and make the path parameter optional, so it works out of the box.

I included an MSYS/GNU build in the Appveyor configuration mainly as an example. Not sure if you care about it.

You'll need to change the badge paths once you've merged this and set up Travis CI and Appveyor appropriately.

Here's a [happy Travis build](https://travis-ci.org/jdub/pathfinder/builds/217462230), and a [happy Appveyor build](https://ci.appveyor.com/project/jdub/pathfinder/build/1.0.26). Enjoy!

pcwalton:
Sorry this slipped through the cracks. :(

This a PR against Pathfinder 1, which is obsolete and has been removed.

PR #32, opened 2017-09-19
demo/client: Make WASD composable

Manishearth:
Currently movement works in the direction of the latest key pressed.
Moving diagonally is not possible, and if you are pressing two keys
and release one, movement stops instead of switching over.

This keeps track of the key press state and updates the movement delta
accordingly.

Note than onKeyDown is called multiple times when you press and hold,
so we must keep track of this state ourselves.

### DIFF
--- a/demo/client/src/camera.ts
+++ b/demo/client/src/camera.ts
@@ -42,6 +42,8 @@ const PERSPECTIVE_INITIAL_ROTATION: glmatrix.vec2 = glmatrix.vec2.clone([Math.PI
 const PERSPECTIVE_OUTER_COLLISION_EXTENT: number = 3000.0;
 const PERSPECTIVE_HITBOX_RADIUS: number = 1.0;
 
+const KEYCODES = ["W", "A", "S", "D"].map(x => x.charCodeAt(0));
+
 export interface OrthographicCameraOptions {
     minScale?: number;
     maxScale?: number;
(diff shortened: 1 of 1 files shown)


demo/client/src/camera.ts
```diff
@@ -42,6 +42,8 @@ const PERSPECTIVE_INITIAL_ROTATION: glmatrix.vec2 = glmatrix.vec2.clone([Math.PI
 const PERSPECTIVE_OUTER_COLLISION_EXTENT: number = 3000.0;
 const PERSPECTIVE_HITBOX_RADIUS: number = 1.0;
 
+const KEYCODES = ["W", "A", "S", "D"].map((x) => x.charCodeAt(0));
```
pcwalton: nit: no need for parens around `(x)`

demo/client/src/camera.ts
```diff
@@ -287,14 +300,37 @@ export class PerspectiveCamera extends Camera {
...
+           }
+        }
+    }
+
+    // Updates the movementDelta vector based on what keys are currently pressed
+    // Returns true if the vector is now empty
+    private updateMovementDelta(): boolean {
+        this.movementDelta = glmatrix.vec3.create();
+        let empty = true;
+        for (let key of KEYCODES) {
```
pcwalton: nit: this can be `const`

PR #37, opened 2017-10-21
Fix compilation on non-macos

Bobo1239:


pcwalton:
Oops, I had worried I messed this up. Thanks!

PR #38, opened 2017-10-25
Workaround to fix compiler issue in TS 2.4

mbebenita:
See: https://github.com/Microsoft/TypeScript/issues/16509

```
ERROR in ./src/svg-demo.ts
(195,61): error TS2345: Argument of type 'Partial<Timings>' is not assignable to parameter of type 'Timings'.
  Types of property 'compositing' are incompatible.
    Type 'number | undefined' is not assignable to type 'number'.
      Type 'undefined' is not assignable to type 'number'.
```

### DIFF
--- a/demo/client/src/3d-demo.ts
+++ b/demo/client/src/3d-demo.ts
@@ -524,7 +524,7 @@ class ThreeDRenderer extends Renderer {
     }
 
     protected newTimingsReceived(): void {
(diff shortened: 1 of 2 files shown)


pcwalton:
I incorporated a fix for this as part of the most recent DirectWrite changes. Thanks for the PR though :)

PR #40, opened 2017-10-25
Zoom Pulse Mode

mbebenita:
Depends on #38 

This PR adds a Zoom Pulse button to slowly zoom in and out the Text and SVG demo. The goal is to demonstrate how smooth rendering is.

image

### DIFF
--- a/demo/client/src/camera.ts
+++ b/demo/client/src/camera.ts
@@ -83,6 +83,7 @@ export abstract class Camera {
         this.canvas = canvas;
     }
 
+    abstract zoom(scale: number): void;
     abstract zoomIn(): void;
     abstract zoomOut(): void;
 }
@@ -148,7 +149,7 @@ export class OrthographicCamera extends Camera {
         glmatrix.vec2.scale(mouseLocation, mouseLocation, window.devicePixelRatio);
 
         const scale = 1.0 - event.deltaY * window.devicePixelRatio * ORTHOGRAPHIC_ZOOM_SPEED;
-        this.zoom(scale, mouseLocation);
+        this._zoom(scale, mouseLocation);
     }
 
     zoomToFit(): void {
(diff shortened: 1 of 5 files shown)


mbebenita:
I should mention that I couldn't find a good place to clear the `setInterval`, so the timer will keep going after the demo is unmounted. I'm not familiar with this UI toolkit, ... so ideas welcome.

trishume:
This has the risk of the zooming looking choppy even if the underlying rendering isn't, because of the timer being de-synchronized from the frames. Probably better to use `requestAnimationFrame` for the animation, this may also solve your issue with unregistering the timer.

demo/client/src/view.ts
```diff
@@ -75,6 +77,22 @@ export abstract class PathfinderView {
...
 
+    zoomPulse(): void {
+        if (this.pulseTimer) {
+            window.clearInterval(this.pulseTimer);
+            this.pulseTimer = 0;
+            return;
+        }
+        let c = 0;
+        let d = 0.005;
+        this.pulseTimer = window.setInterval(() => {
```
pcwalton: Please use `requestAnimationFrame()`

demo/client/src/camera.ts
```diff
@@ -306,6 +311,10 @@ export class PerspectiveCamera extends Camera {
         ]);
     }
 
+    zoom(scale: number): void {
```
pcwalton: Why this change?

mbebenita: I wanted to keep the old `zoom` function private, but I needed something public that only took the `scale` argument. So I renamed the private function to `_zoom` and exposed the new function as `zoom`.

demo/client/src/view.ts
mbebenita: Will do.

mbebenita:
@trishume good idea, I switched to RAFs in the latest patch.

pcwalton:
Works for me.

PR #47, opened 2017-11-09
Run demo on new typescript

dikaiosune:
Not a particularly useful PR, but when attempting to run the demo locally the semver bound in package.json allows TypeScript 2.6.1 to be installed, which seems to break the client demo build:

```
ERROR in ./src/text.ts
(251,23): error TS2345: Argument of type '{ body: string; headers: { 'Content-Type': string; }; method: string;}' is not assignable to parameter of type 'RequestInit | undefined'.
  Type '{ body: string; headers: { 'Content-Type': string; }; method: string; }' is not assignable to type 'RequestInit'.
    Types of property 'headers' are incompatible.
      Type '{ 'Content-Type': string; }' is not assignable to type 'Headers | string[][] | undefined'.
        Object literal may only specify known properties, and ''Content-Type'' does not exist in type 'Headers | string[][] | undefined'.

ERROR in ./src/text.ts
(253,17): error TS7006: Parameter 'response' implicitly has an 'any' type.

ERROR in ./src/text.ts
(256,17): error TS7006: Parameter 'buffer' implicitly has an 'any' type.

ERROR in ./src/svg-loader.ts
(110,23): error TS2345: Argument of type '{ body: string; headers: { 'Content-Type': string; }; method: string;}' is not assignable to parameter of type 'RequestInit | undefined'.
  Type '{ body: string; headers: { 'Content-Type': string; }; method: string; }' is not assignable to type 'RequestInit'.
    Types of property 'headers' are incompatible.
      Type '{ 'Content-Type': string; }' is not assignable to type 'Headers | string[][] | undefined'.
        Object literal may only specify known properties, and ''Content-Type'' does not exist in type 'Headers | string[][] | undefined'.

ERROR in ./src/svg-loader.ts
(112,17): error TS7006: Parameter 'response' implicitly has an 'any' type.

ERROR in ./src/svg-loader.ts
(115,17): error TS7006: Parameter 'buffer' implicitly has an 'any' type.
```

I also added yarn.lock to gitignore since the README recommends npm, but would be happy to revert.

I suspect that resolving the TS errors with more accurate type defs might be more reasonable, but this allowed me to run the demo on a fresh clone.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ Cargo.lock
 
 # VSCode
 .vscode
+
+# this project recommends npm
+yarn.lock
--- a/demo/client/src/svg-loader.ts
+++ b/demo/client/src/svg-loader.ts
@@ -107,7 +107,7 @@ export class SVGLoader {
         let time = 0;
         return window.fetch(PARTITION_SVG_PATHS_ENDPOINT_URL, {
             body: JSON.stringify({ paths: paths }),
-            headers: {'Content-Type': 'application/json'},
+            headers: [['Content-Type', 'application/json']],
             method: 'POST',
         }).then(response => {
             time = parseServerTiming(response.headers);
--- a/demo/client/src/text.ts
+++ b/demo/client/src/text.ts
@@ -248,7 +248,7 @@ export class GlyphStore {
         let time = 0;
         return window.fetch(PARTITION_FONT_ENDPOINT_URI, {
             body: JSON.stringify(request),
-            headers: {'Content-Type': 'application/json'},
+            headers: [['Content-Type', 'application/json']],
             method: 'POST',
         }).then(response => {
             time = parseServerTiming(response.headers);


Bobo1239:
This is https://github.com/Microsoft/TypeScript/issues/19656 whose [fix](https://github.com/Microsoft/TypeScript/pull/19690) isn't in a release yet. If you switch typescript's version to `next` in `package.json` it builds without errors.

dikaiosune:
Oh, fantastic!

whmountains:
Don't close, we need the fix.

dikaiosune:
Really? Presumably the TS changes will release soon?

pcwalton:
I'm confused :)

dikaiosune:
from irc if anyone's following along here:

```
4:14 PM <dikaiosune> pcwalton: so am i
4:15 PM <•pcwalton> dikaiosune: I’m going to keep the PR open in case MS releases a new version soon
4:15 PM <•pcwalton> if not, then I’ll merge it
```

pcwalton:
2578298198c52ab1dbbb49e3859b4144327b7aeb includes this.

PR #55, opened 2017-12-06
Fixes #51 by introducing a variable for an early return boolean.

### LINKED ISSUE #51: MacBook Pro 2014 fails to render XCAA text (Firefox+Chrome)
I originally filed #2 for Pathfinder 1, but it appears that Pathfinder 2 is just as broken on my 2014 Macbook Pro.

I've tested under both Firefox and Chrome and neither can successfully render glyphs in XCAA mode, but work just fine in 4xSSAA.

System

  Model Name:	MacBook Pro
  Model Identifier:	MacBookPro11,3
  Processor Name:	Intel Core i7
  Processor Speed:	2.8 GHz

GPUs:

1

  Chipset Model:	Intel Iris Pro
  Type:	GPU
  Bus:	Built-In
  VRAM (Dynamic, Max):	1536 MB
  Vendor:	Intel
  Device ID:	0x0d26
  Revision ID:	0x0008
  Automatic Graphics Switching:	Supported
  gMux Version:	4.0.8 [3.2.8]
  Metal:	Supported, feature set macOS GPUFamily1 v3

2

  Chipset Model:	NVIDIA GeForce GT 750M
  Type:	GPU
  Bus:	PCIe
  PCIe Lane Width:	x8
  VRAM (Total):	2 GB
  Vendor:	NVIDIA (0x10de)
  Device ID:	0x0fe9
  Revision ID:	0x00a2
  ROM Revision:	3776
  Automatic Graphics Switching:	Supported
  gMux Version:	4.0.8 [3.2.8]
  Metal:	Supported, feature set macOS GPUFamily1 v3

screen shot 2017-11-27 at 10 34 10 am
screen shot 2017-11-27 at 10 34 19 am

trishume:
See #51. This works around a bug in the macOS Nvidia shader compiler in 2014 rMBPs. The fix shows that as suspected this is very clearly a GPU bug. The funny part is that the compiler is clearly not great since putting the condition in a variable that isn't used anywhere shouldn't make much difference to a competent compiler.

I hope putting the condition in a variable doesn't significantly regress performance on other platforms. Might be worth a benchmark.

I also removed the unused `lineDoesNotPassThroughPixel` function.

cc @pcwalton @mmastrac

### DIFF
--- a/shaders/gles2/common.inc.glsl
+++ b/shaders/gles2/common.inc.glsl
@@ -358,10 +358,6 @@ vec4 clipLineToPixelRow(vec2 p0, vec2 dP, float pixelCenterY, out float outPixel
     return vec4(p0 + dP * tY.x, dP * (tY.y - tY.x));
 }
 
-bool lineDoesNotPassThroughPixel(vec2 t) {
-    return t.x >= t.y;
-}
-
 /// Computes the area of the polygon covering the pixel with the given boundaries.
 ///
 /// The line must run left-to-right and must already be clipped to the left and right sides of the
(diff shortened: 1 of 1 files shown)


leeoniya:
this looks like a new one.

might be worth adding to https://github.com/servo/webrender/wiki/Driver-issues

trishume:
@leeoniya Done!

pcwalton:
UGH. Thanks for tracking this down!

PR #74, opened 2018-03-09
The Send impl for `FontContext` is too forgiving.

pythonesque:
I'm pretty confident that it is only thread safe if FK is (or another way of putting it is, I'm pretty sure if you include an `Rc<_>` in FK it should not be safe to share between threads).


pcwalton:
Good catch!

PR #75, opened 2018-03-09
Super hacky webvr support

Manishearth:
don't merge this, it breaks a bunch of things.

To use:

 - Open the 3d demo in Chrome Canary for Android with webvr enabled (chrome://flags)
 - Select an AA strategy _other_ than 4xSSAA (this hits memory limits and needs investigation)
 - Click the VR button _twice_ . This is due to a bug in how chrome determines if a VR request was initiated by user action
 - Put the phone in your Daydream and press the home key on your controller
 - Connect a physical keyboard to navigate. Yes, really.

The AA breaks down at a distance and shows fuzzy stuff, so you'll need to move closer to the monument to see text.

Will work on a proper mergeable PR later.

### DIFF
--- a/demo/client/html/3d-demo.html.hbs
+++ b/demo/client/html/3d-demo.html.hbs
@@ -42,6 +42,12 @@
                  aria-expanded="false" aria-controls="#pf-settings">
                     <span class="pf-material-icons">settings</span>
                 </button>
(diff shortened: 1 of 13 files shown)


Manishearth:
https://github.com/pcwalton/pathfinder/pull/76

PR #76, opened 2018-03-10
Cleaned up VR implementation

Manishearth:
Actually mergeable.

I still want to improve the camera code. It currently just overwrites the rotation matrix, but we can do better.

Ideally the code gets refactored into having two view transforms.

r? @pcwalton

### DIFF
--- a/demo/client/html/3d-demo.html.hbs
+++ b/demo/client/html/3d-demo.html.hbs
@@ -46,7 +46,10 @@
                  class="btn btn-outline-secondary pf-toolbar-button"
                  aria-expanded="false" aria-controls="#pf-vr"
                  style="display: none">
-                    <span>VR</span>
+                    <!-- https://materialdesignicons.com/icon/virtual-reality -->
+                    <svg style="height: 18px" viewBox="0 0 24 24">
+                        <path fill="#000000" d="M5,3C3.89,3 3,3.9 3,5V19A2,2 0 0,0 5,21H19A2,2 0 0,0 21,19V5A2,2 0 0,0 19,3H5M6,9H7.5L8.5,12.43L9.5,9H11L9.25,15H7.75L6,9M13,9H16.5C17.35,9 18,9.65 18,10.5V11.5C18,12.1 17.6,12.65 17.1,12.9L18,15H16.5L15.65,13H14.5V15H13V9M14.5,10.5V11.5H16.5V10.5H14.5Z" />
+                    </svg>
                 </button>
             </div>
         </div>
--- a/demo/client/html/3d-demo.html.hbs
+++ b/demo/client/html/3d-demo.html.hbs
@@ -42,6 +42,12 @@
                  aria-expanded="false" aria-controls="#pf-settings">
                     <span class="pf-material-icons">settings</span>
                 </button>
+                <button id="pf-vr" type="button"
+                 class="btn btn-outline-secondary pf-toolbar-button"
+                 aria-expanded="false" aria-controls="#pf-vr"
+                 style="display: none">
+                    <span>VR</span>
+                </button>
             </div>
         </div>
         {{>partials/spinner.html}}
--- a/demo/client/src/3d-demo.ts
+++ b/demo/client/src/3d-demo.ts
@@ -719,7 +719,7 @@ class ThreeDRenderer extends Renderer {
                          normal[0] / normal[3],
                          normal[1] / normal[3],
                          normal[2] / normal[3]);
-
+            gl.uniform1f(monumentProgram.uniforms.uLightThings, this.inVR ? 0 : 1);
             // Draw the face!
             gl.drawElements(gl.TRIANGLES, 6, gl.UNSIGNED_SHORT, face * 6 * UINT16_SIZE);
         }
--- a/demo/client/src/3d-demo.ts
+++ b/demo/client/src/3d-demo.ts
@@ -370,7 +370,7 @@ class ThreeDView extends DemoView implements TextRenderContext {
         this.inVRRAF = false;
         this.vrDisplay = null;
         if ("VRFrameData" in window) {
-           this.vrFrameData = new VRFrameData;
+            this.vrFrameData = new VRFrameData;
         } else {
             this.vrFrameData = null;
         }
@@ -383,44 +383,42 @@ class ThreeDView extends DemoView implements TextRenderContext {
     newTimingsReceived(timings: Timings): void {}
 
     vrSetup(): void {
-        if (navigator.getVRDisplays) {
-            navigator.getVRDisplays().then((displays) => {
-              if (displays.length > 0) {
-                this.vrDisplay = displays[displays.length - 1];
-
-                // It's heighly reccommended that you set the near and far planes to
-                // something appropriate for your scene so the projection matricies
-                // WebVR produces have a well scaled depth buffer.
-                this.renderer.setClipPlanes(this.vrDisplay);
-                unwrapNull(document.getElementById('pf-vr')).style.display = "initial";
-              } else {
-                // no vr displays
-              }
-            });
-        } else {
-            // no vr support
+        if (!('getVRDisplays' in navigator)) {
+            return;
         }
+
+        navigator.getVRDisplays().then(displays => {
+            if (displays.length === 0) {
+                return;
+            }
+            this.vrDisplay = displays[displays.length - 1];
+
+            // It's heighly recommended that you set the near and far planes to
+            // something appropriate for your scene so the projection matrices
+            // WebVR produces have a well scaled depth buffer.
+            this.renderer.setClipPlanes(this.vrDisplay);
+            unwrapNull(document.getElementById('pf-vr')).style.display = "initial";
+        });
 
         window.addEventListener('vrdisplaypresentchange', () => {
           if (this.vrDisplay == null)
               return;
 
           if (this.vrDisplay.isPresenting) {
-            const that = this;
             const eye = this.vrDisplay.getEyeParameters("left");
 
             this.vrDisplayHeight = eye.renderHeight;
             this.vrDisplayWidth = eye.renderWidth * 2;
             this.resizeToFit(true);
-            function vrCallback(): void {
-                if (that.vrDisplay == null || !that.renderer.inVR) {
+            const vrCallback = () => {
+                if (this.vrDisplay == null || !this.renderer.inVR) {
                     return;
                 }
-                that.vrDisplay.requestAnimationFrame(vrCallback);
-                that.inVRRAF = true;
-                that.redraw();
-                that.inVRRAF = false;
-            }
+                this.vrDisplay.requestAnimationFrame(vrCallback);
+                this.inVRRAF = true;
+                this.redraw();
+                this.inVRRAF = false;
+            };
             this.vrDisplay.requestAnimationFrame(vrCallback);
           } else {
             this.renderer.inVR = false;
@@ -458,7 +456,7 @@ class ThreeDRenderer extends Renderer {
     camera: PerspectiveCamera;
 
     needsStencil: boolean = false;
-    rightEye: boolean = false;
+    vrRightEye: boolean = false;
 
     get isMulticolor(): boolean {
         return false;
@@ -532,6 +530,7 @@ class ThreeDRenderer extends Renderer {
 
         this.distantGlyphVAO = null;
         this.vrProjectionMatrix = null;
+
         this.camera = new PerspectiveCamera(renderContext.canvas, {
             innerCollisionExtent: MONUMENT_SCALE[0],
         });
@@ -573,10 +572,10 @@ class ThreeDRenderer extends Renderer {
     redrawVR(frame: VRFrameData): void {
         this.clearDestFramebuffer(true);
         this.vrProjectionMatrix = frame.leftProjectionMatrix;
-        this.rightEye = false;
+        this.vrRightEye = false;
         this.camera.setView(F32ArrayToMat4(frame.leftViewMatrix), frame.pose);
         this.redraw();
-        this.rightEye = true;
+        this.vrRightEye = true;
         this.vrProjectionMatrix = frame.rightProjectionMatrix;
         this.camera.setView(F32ArrayToMat4(frame.rightViewMatrix), frame.pose);
         this.redraw();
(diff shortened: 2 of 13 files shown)


demo/client/src/renderer.ts
```diff
@@ -146,7 +146,9 @@ export abstract class Renderer {
         this.clearDestFramebuffer();
 
         // Start timing rendering.
-        if (this.timerQueryPollInterval == null) {
+        if (this.timerQueryPollInterval == null &&
```
pcwalton: The entire commit that makes timer queries optional shouldn't be necessary now since I made this change right before showing my slides at the Rust meetup.

demo/client/html/3d-demo.html.hbs
```diff
@@ -42,6 +42,12 @@
                  aria-expanded="false" aria-controls="#pf-settings">
                     <span class="pf-material-icons">settings</span>
                 </button>
+                <button id="pf-vr" type="button"
+                 class="btn btn-outline-secondary pf-toolbar-button"
+                 aria-expanded="false" aria-controls="#pf-vr"
+                 style="display: none">
+                    <span>VR</span>
```
pcwalton: Note to self: we use Material Design Icons here and there is a VR icon, so let's use that.

demo/client/src/view.ts
```diff
@@ -174,6 +175,10 @@ export abstract class DemoView extends PathfinderView implements RenderContext {
     protected colorBufferHalfFloatExt: any;
 
     private wantsScreenshot: boolean;
+    private vrDisplay: VRDisplay | null;
```
pcwalton: nit: can you put a newline above this? Let's keep all the VR-specific stuff together.

demo/client/src/view.ts
```diff
@@ -197,6 +202,60 @@ export abstract class DemoView extends PathfinderView implements RenderContext {
...
         this.atlasRenderingTimerQuery = null;
         this.compositingTimerQuery = null;
+
+        this.inVRRAF = false;
+        this.inVR = false;
+        this.vrDisplay = null;
+        if ("VRFrameData" in window) {
+           this.vrFrameData = new VRFrameData;
+        } else {
+            this.vrFrameData = null;
```
pcwalton: nit: weird indentation here

demo/client/src/view.ts
```diff
@@ -197,6 +202,60 @@ export abstract class DemoView extends PathfinderView implements RenderContext {
...
+        }
+
+        this.vrSetup();
+    }
+
+    vrSetup(): void {
+        if (navigator.getVRDisplays) {
```
pcwalton: nit: Let's use `if ('getVRDisplays' in navigator) { ... }` for consistency with the above.

demo/client/src/view.ts
```diff
@@ -197,6 +202,60 @@ export abstract class DemoView extends PathfinderView implements RenderContext {
...
+            navigator.getVRDisplays().then((displays) => {
```
pcwalton: nit: no need for parens around `displays`

demo/client/src/view.ts
```diff
@@ -197,6 +202,60 @@ export abstract class DemoView extends PathfinderView implements RenderContext {
...
+              if (displays.length > 0) {
```
pcwalton: Usually I use early-return to reduce rightward drift, so I'd write this as:

    if (displays.length === 0)
        return;

demo/client/src/view.ts
```diff
@@ -197,6 +202,60 @@ export abstract class DemoView extends PathfinderView implements RenderContext {
...
+                // something appropriate for your scene so the projection matricies
+                // WebVR produces have a well scaled depth buffer.
+                this.vrDisplay.depthNear = NEAR_CLIP_PLANE;
+                this.vrDisplay.depthFar = FAR_CLIP_PLANE;
+                unwrapNull(document.getElementById('pf-vr')).style.display = "initial";
+              } else {
+                // no vr displays
+              }
+            });
+        } else {
```
pcwalton: As above, let's use early-return to reduce rightward drift.

demo/client/src/view.ts
```diff
@@ -197,6 +202,60 @@ export abstract class DemoView extends PathfinderView implements RenderContext {
...
+        }
+
+        window.addEventListener('vrdisplaypresentchange', () => {
+
+          if (this.vrDisplay == null)
+              return;
+
+          if (this.vrDisplay.isPresenting) {
+            const that = this;
+            function vrCallback(): void {
```
pcwalton: Can't we use an arrow function here to get lexical `this`? `const vrCallback = () => { ... }`

demo/client/src/view.ts
```diff
@@ -197,6 +202,60 @@ export abstract class DemoView extends PathfinderView implements RenderContext {
...
+        this.vrSetup();
+    }
+
+    vrSetup(): void {
+        if (navigator.getVRDisplays) {
+            navigator.getVRDisplays().then((displays) => {
+              if (displays.length > 0) {
+                this.vrDisplay = displays[displays.length - 1];
+
+                // It's heighly reccommended that you set the near and far planes to
```
pcwalton: typo: should be "recommended"

demo/client/src/view.ts
```diff
@@ -197,6 +202,60 @@ export abstract class DemoView extends PathfinderView implements RenderContext {
...
+                // something appropriate for your scene so the projection matricies
```
pcwalton: typo: should be "matrices"

demo/client/src/3d-demo.ts
```diff
@@ -124,6 +124,15 @@ interface MeshDescriptor {
     positions: glmatrix.vec2[];
 }
 
+function F32ArrayToMat4(array: Float32Array): mat4 {
+    const mat = glmatrix.mat4.create();
```
pcwalton: I believe you can just remove this function and use `glmatrix.mat4.clone(array)` instead: http://glmatrix.net/docs/mat4.js.html#line60

demo/client/src/3d-demo.ts
```diff
@@ -433,7 +444,7 @@ class ThreeDRenderer extends Renderer {
         this.glyphSizes = [];
 
         this.distantGlyphVAO = null;
-
+        this.vrProjectionMatrix = null;
```
pcwalton: nit: put a newline after this to keep VR stuff separated out

demo/client/src/3d-demo.ts
```diff
@@ -472,6 +483,15 @@ class ThreeDRenderer extends Renderer {
         this.renderContext.gl.uniform4f(uniforms.uHints, 0, 0, 0, 0);
     }
 
+    redrawVR(frame: VRFrameData): void {
+        this.vrProjectionMatrix = frame.leftProjectionMatrix;
+        this.camera.setView(F32ArrayToMat4(frame.leftViewMatrix), frame.pose);
+        this.redraw();
+        this.vrProjectionMatrix = frame.rightProjectionMatrix;
+        this.camera.setView(F32ArrayToMat4(frame.rightViewMatrix), frame.pose);
+        this.redraw();
```
pcwalton: Could you add a comment here specifying why you need to redraw twice? Also I believe you can replace `F32ArrayToMat4` with `glmatrix.mat4.clone()`, as above.

demo/client/src/3d-demo.ts
```diff
@@ -811,6 +831,9 @@ class ThreeDRenderer extends Renderer {
     }
 
     private calculateProjectionTransform(): glmatrix.mat4 {
+        if (this.vrProjectionMatrix != null) {
+            return F32ArrayToMat4(this.vrProjectionMatrix);
```
pcwalton: ditto, `glmatrix.mat4.clone`.

demo/client/src/camera.ts
```diff
@@ -370,6 +373,10 @@ export class PerspectiveCamera extends Camera {
         // TODO(pcwalton)
     }
 
+    setView(rotation: glmatrix.mat4, pose: VRPose) {
```
pcwalton: nit: annotate the return type with `: void`

demo/client/src/3d-demo.ts
```diff
@@ -376,6 +376,7 @@ class ThreeDRenderer extends Renderer {
     camera: PerspectiveCamera;
 
     needsStencil: boolean = false;
+    rightEye: boolean = false;
```
pcwalton: nit: put a newline above this. Also maybe call this `vrRightEye` to make it clear that this is for VR?

demo/client/src/3d-demo.ts
```diff
@@ -484,14 +489,27 @@ class ThreeDRenderer extends Renderer {
...
         this.vrProjectionMatrix = frame.rightProjectionMatrix;
         this.camera.setView(F32ArrayToMat4(frame.rightViewMatrix), frame.pose);
         this.redraw();
     }
 
+    setDrawViewport() {
+        let offset = 0;
+        if (this.rightEye) {
+            offset = this.destAllocatedSize[0];
+        }
```
pcwalton: nit: let's just write `const offset = this.vrRightEye ? this.destAllocatedSize[0] : 0;` in order to make `offset` immutable.

demo/client/src/3d-demo.ts
```diff
@@ -484,14 +489,27 @@ class ThreeDRenderer extends Renderer {
...
+        this.rightEye = false;
         this.camera.setView(F32ArrayToMat4(frame.leftViewMatrix), frame.pose);
         this.redraw();
+        this.rightEye = true;
         this.vrProjectionMatrix = frame.rightProjectionMatrix;
         this.camera.setView(F32ArrayToMat4(frame.rightViewMatrix), frame.pose);
         this.redraw();
     }
 
+    setDrawViewport() {
```
pcwalton: nit: annotate the return type with `: void`

demo/client/src/renderer.ts
```diff
@@ -141,13 +143,19 @@ export abstract class Renderer {
         this.redraw();
     }
 
+    setDrawViewport() {
```
pcwalton: nit: as above, annotate the return type with `: void`

shaders/gles2/demo-3d-monument.fs.glsl
```diff
@@ -25,22 +25,28 @@ uniform float uShininess;
 /// The normal of these vertices.
 uniform vec3 uNormal;
 
+uniform bool uLightThings;
```
pcwalton: Cute name, but I think that `uEnableLighting` might be more obvious :)

demo/client/src/3d-demo.ts
```diff
@@ -364,10 +367,85 @@ class ThreeDView extends DemoView implements TextRenderContext {
         this.appController = appController;
         this.renderer = new ThreeDRenderer(this);
 
+        this.inVRRAF = false;
+        this.vrDisplay = null;
+        if ("VRFrameData" in window) {
+           this.vrFrameData = new VRFrameData;
+        } else {
+            this.vrFrameData = null;
```
pcwalton: nit: weird indentation

demo/client/src/3d-demo.ts
```diff
@@ -364,10 +367,85 @@ class ThreeDView extends DemoView implements TextRenderContext {
...
+                // WebVR produces have a well scaled depth buffer.
+                this.renderer.setClipPlanes(this.vrDisplay);
+                unwrapNull(document.getElementById('pf-vr')).style.display = "initial";
+              } else {
+                // no vr displays
+              }
+            });
+        } else {
+            // no vr support
+        }
```
pcwalton: Stylistic comments left on the earlier version of this function apply here too.

demo/client/src/renderer.ts
```diff
@@ -149,6 +149,9 @@ export abstract class Renderer {
         gl.viewport(0, 0, this.destAllocatedSize[0], this.destAllocatedSize[1]);
     }
 
+    setClipPlanes(display: VRDisplay) {
```
pcwalton: Why this function? (Also needs a return type annotation.)

demo/client/src/3d-demo.ts
```diff
@@ -406,21 +406,20 @@ class ThreeDView extends DemoView implements TextRenderContext {
...
-                that.vrDisplay.requestAnimationFrame(vrCallback);
-                that.inVRRAF = true;
-                that.redraw();
-                that.inVRRAF = false;
-            }
+                this.vrDisplay.requestAnimationFrame(vrCallback);
+                this.inVRRAF = true;
+                this.redraw();
+                this.inVRRAF = false;
+            };
```
pcwalton: Ah, I see you made the arrow function change. Disregard the previous comments then :)

demo/client/html/3d-demo.html.hbs
Manishearth: huh, I looked for one and didn't find it. Will look again

Manishearth: https://github.com/google/material-design-icons/issues/508

demo/client/src/view.ts
Manishearth: lol this is a comment from the demo app, I should fix that

Manishearth:
addressed

pcwalton:
@Manishearth I found a community icon here: https://materialdesignicons.com/icon/virtual-reality

pcwalton:
(Note that this still has merge conflicts just as a reminder)

Manishearth:
pushed. I'll fix the icon later, seems to be trickier since it's not in the font.

Manishearth:
Okay -- so I looked into it; that icon is only in the fonts provided on https://materialdesignicons.com/getting-started , however these work a completely different way; they don't work by ligaturifying the right text. Instead, they use a bunch of CSS with ::before attributes being used with Private Use codepoints.


I'm not sure if we should switch to that model, seems tricky to style.

pcwalton:
We could always just embed the SVG.

On Thu, Mar 29, 2018 at 5:15 AM, Manish Goregaokar <notifications@github.com

Manishearth:
Pushed. It's hard to get it at the same size, but it's close enough.

pcwalton:
Looks good, squash into a few commits and let's get it merged :)

Manishearth:
Squashed the fixups, kept the main commits so that it's easy to read.

pcwalton:
Merged!

PR #70, opened 2018-02-24
chore: use freetype to replace freetype-sys

Brooooooklyn:
I'm trying to build a node canvas with rust: [context_2d](https://github.com/rust-canvas/rust-canvas/blob/feature/font-renderer/src/canvas/context_2d.rs#L121), and borrowed tones of codes from [servo](https://github.com/servo/servo/blob/master/components/canvas/canvas_paint_thread.rs).

When I was trying to built it on linux, it crashed with:

```
error: multiple packages link to native library `freetype`, but a native library can be linked only once

package `servo-freetype-sys v4.0.3`
    ... which is depended on by `azure v0.25.0`
    ... which is depended on by `rustcanvas v0.1.0 (file:///Users/longyinan/workspace/Github/rust-canvas)`
links to native library `freetype`

package `freetype-sys v0.6.0`
    ... which is depended on by `pathfinder_font_renderer v0.1.0 (https://github.com/rust-canvas/pathfinder?branch=rust-canvas#5f082aa4)`
    ... which is depended on by `rustcanvas v0.1.0 (file:///Users/longyinan/workspace/Github/rust-canvas)`
also links to native library `freetype`
```

So I change the dependencies in pathfinder to pass the compile, and I have no idea about the differences between the `freetype-sys` & `freetype`.

Please tell me if you think this Pull Request is unreasonable, and feel free to close it.

PR #78, opened 2018-04-05
Use servo-freetype instead of the system-native freetype

fschutt:
See https://github.com/servo/webrender/issues/2617

It seems that `freetype-sys`, which is what pathfinder uses, links to the system-native freetype version, i.e. whatever you happen to have installed right now. webrender, on the other hand, uses `freetype`, which is the `servo-freetype` version, which embeds a full copy of freetype into the binary. This also means that pathfinder shouldn't build on linux if you don't have freetype installed.

I had to rename a few things, because a feature, a module and a crate that are all named "freetype" led to naming conflicts.

### DIFF
--- a/font-renderer/Cargo.toml
+++ b/font-renderer/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Patrick Walton <pcwalton@mimiga.net>"]
 
 [features]
 default = []
-freetype = ["freetype-sys"]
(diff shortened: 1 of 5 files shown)


Brooooooklyn:
see https://github.com/pcwalton/pathfinder/pull/70

pcwalton:
Just did this myself. Thanks!

PR #81, opened 2018-04-13
Fix the API of glyph_dimensions to make webrender compile on Windows

fschutt:
See https://github.com/servo/webrender/issues/2645 - this fixes 2 of the 6 errors I get when compiling webrender on Windows. 

I decided to use the same API that is used for Mac & Linux builds, with the extra `exact` parameter and returning an Error instead of an Option. Maybe it would be a good idea to change it back to `Option` instead of `Err<()>`, but then please change it in all 3 APIs.


fschutt:
The Windows `PathfinderFontContext` was missing the `add_native_font` function. I've simply copied + adapted the function from the Mac version + exported the necessary types (so that in webrender we can have a `Into<PathfinderComPtr<IDWriteFontFace>>` (similar to the current Mac interface).

fschutt:
@pcwalton I made the API more or less like the Mac version, I hope it's ok to do this.

pcwalton:
Merged, sorry for the delay! Thanks!

PR #82, opened 2018-05-03
Fixed calculation of glyph dimensions on Windows

fschutt:
The dimensions in the `DWRITE_GLYPH_METRICS` are a bit special: the `rightSideBearing` is measured from the right edge of the character boundary, not the left edge.

So, doing (right - left) would usually return a negative or very small number, which, when casted to a u32 in an unsafe block, would silently overflow to a very large number and make webrender crash.

Therefore, in order to calculate the metrics correctly, we have to first calculate the full width of the character, then subtract the right and left edges, same width the height.


pcwalton:
Merged, thanks!

PR #89, opened 2018-07-10
Update lyon and euclid dependencies

nical:
Needed for https://github.com/servo/webrender/pull/2882


nical:
cc @BorisChiou

nical:
Note that lyon crates reexpose their public dependencies so I changed most of the code that depend on both lyon_geom and lyon_path to only depend only on lyon_path. That way there's a little less work to do to update the two (new versions of the lyon crates are always published at the same time). lyon_geom also reexports euclid but I didn't touch that in most places as pathfinder uses certain euclid feature flags.

BorisChiou:
Awesome! Thanks for this patch.

nox:
@pcwalton Can we move this repository to the Servo organisation? Thanks.

PR #104, opened 2018-12-23
Run rustfmt in Travis

### LINKED ISSUE #79: Run rustfmt
We should automatically format code so I don't have to annoy contributors with style nits.

CYBAI:
This PR will add Travis to run `build` and also `fmt` so that it should fix #79 .

Contributors can run `rustfmt **/*.rs` to fix the styles.

Besides, @pcwalton, would you prefer to use rustfmt config as same as [the Servo one](https://github.com/servo/servo/blob/master/rustfmt.toml)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: rust
+rust:
+  - nightly
+  - beta
+  - stable
+script: |
+  cargo fmt -- --check &&
+  cargo build --verbose &&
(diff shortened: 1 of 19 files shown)


pcwalton:
This is no longer applicable to Pathfinder 3. Sorry for dropping the ball here :(

I think we should use the plain old default `rustfmt` formatting, not Servo's.

PR #109, opened 2019-03-11
Get pf-no-simd to compile on aarch64

asajeffrey:
This gets the simd crate to compile for aarch64 with the `pf-no-simd` feature.


pcwalton:
I already have these changes locally, sorry :)

I'm leaving this open until I push them.

asajeffrey:
Fixed by #110.

PR #111, opened 2019-03-20
Add Transform3DF32::from_quaternion

asajeffrey:
Add a function to build a transform from a quaternion.

### DIFF
--- a/geometry/src/basic/transform3d.rs
+++ b/geometry/src/basic/transform3d.rs
@@ -97,6 +97,24 @@ impl Transform3DF32 {
                                   0.0, 0.0, 0.0, 1.0)
     }
 
+    /// Just like `glm::quaternion::mat4_cast()`
+    #[inline]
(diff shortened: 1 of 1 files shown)


pcwalton:
Needs SIMD

asajeffrey:
Why does this need SIMD but, e.g, `from_ortho` doesn't? If it makes any difference, the way this is going to be used it won't be on a hot path. We could check it in with a `TODO: SIMD?`

pcwalton:
The code will be clearer with SIMD. Give me a few minutes and I'll make a go of it.

pcwalton:
(I'd also prefer to use rotors instead of quaternions, but I'm OK with deferring that.)

pcwalton:
I pushed this patch, partially SIMD-ified: https://github.com/pcwalton/pathfinder/commit/26fea26f87427b1ac179f0ed77c6fc9bbc155b13

(It's untested but I triple checked it.)

Thanks for the patch!

PR #113, opened 2019-03-20
Allow a non-zero default framebuffer

asajeffrey:
Allows using pathfinder when the default framebuffer is non-zero (e.g. on ML1).

### DIFF
--- a/gl/src/lib.rs
+++ b/gl/src/lib.rs
@@ -24,15 +24,18 @@ use std::mem;
 use std::ptr;
 use std::str;
 use std::time::Duration;
-
 pub struct GLDevice {
     version: GLVersion,
+    default_framebuffer: GLuint,
 }
 
 impl GLDevice {
     #[inline]
-    pub fn new(version: GLVersion) -> GLDevice {
-        GLDevice { version }
+    pub fn new(version: GLVersion, default_framebuffer: GLuint) -> GLDevice {
+        GLDevice {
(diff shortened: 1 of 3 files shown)


gl/src/lib.rs
```diff
@@ -27,12 +27,20 @@ use std::time::Duration;
...
+    default_framebuffer_id: GLuint,
 }
 
 impl GLDevice {
     #[inline]
     pub fn new(version: GLVersion) -> GLDevice {
-        GLDevice { version }
+        let mut default_framebuffer_id = 0;
+        unsafe {
+            gl::GetIntegerv(gl::FRAMEBUFFER_BINDING, &mut default_framebuffer_id);
```
pcwalton: This can be slow, and it's hidden global state. I think it's probably better to just pass in the framebuffer as a parameter.

gl/src/lib.rs
```diff
@@ -27,12 +27,20 @@ use std::time::Duration;
 
 pub struct GLDevice {
     version: GLVersion,
+    default_framebuffer_id: GLuint,
```
pcwalton: uber-nit: "object" is the preferred GL terminology over "id"

asajeffrey: I renamed to `default_framebuffer`.

asajeffrey:
@pcwalton review ping?

PR #114, opened 2019-03-20
Allow devices to set transforms for each camera

asajeffrey:
Allow the device to set different camera transforms for each eye.

### DIFF
--- a/demo/common/src/lib.rs
+++ b/demo/common/src/lib.rs
@@ -12,7 +12,7 @@
 
 use crate::device::{GroundLineVertexArray, GroundProgram, GroundSolidVertexArray};
 use crate::ui::{BackgroundColor, DemoUI, UIAction};
-use crate::window::{Event, Keycode, SVGPath, Window, WindowSize};
+use crate::window::{CameraTransform, Event, Keycode, SVGPath, Window, WindowSize};
 use clap::{App, Arg};
 use image::ColorType;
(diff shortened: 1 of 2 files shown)


asajeffrey:
Rebased.

asajeffrey:
Rebased. @pcwalton review ping?

asajeffrey:
Rebased.

demo/common/src/lib.rs
```diff
@@ -923,19 +914,48 @@ fn center_of_window(window_size: &WindowSize) -> Point2DF32 {
 
 enum Camera {
     TwoD(Transform2DF32),
-    ThreeD { transform: CameraTransform3D, velocity: Point3DF32 },
+    ThreeD {
+        // The mode will either be 3D or VR
+        mode: Mode,
```
pcwalton: This seems redundant; the mode can be detected by looking at the length of `transforms`. How about making this a method (e.g. `is_stereo()`) on `Camera`?

asajeffrey: Fixed. I kept the `mode()` method, rather than going for `is_stereo()`, just because that fits the way the method is used.

asajeffrey:
Rebased.

pcwalton:
Merged in 3996d9a653f633742010df6aaff6ddca15f2e492.

PR #115, opened 2019-03-20
Allow devices to determine the viewport for each eye

asajeffrey:
This makes the Window determine the viewports, not the demo appl

### DIFF
--- a/demo/android/rust/src/lib.rs
+++ b/demo/android/rust/src/lib.rs
@@ -17,6 +17,7 @@ use pathfinder_demo::DemoApp;
 use pathfinder_demo::Options;
 use pathfinder_demo::window::{Event, SVGPath, Window, WindowSize};
 use pathfinder_geometry::basic::point::Point2DI32;
+use pathfinder_geometry::basic::rect::RectI32;
 use pathfinder_gl::GLVersion;
 use pathfinder_gpu::resources::ResourceLoader;
 use std::cell::RefCell;
@@ -48,6 +49,7 @@ pub unsafe extern "system" fn
                                                                             height: i32) {
     let logical_size = Point2DI32::new(width, height);
     let window_size = WindowSize { logical_size, backing_scale_factor: 1.0 };
+    let window = WindowImpl { size: logical_size };
 
     JAVA_ACTIVITY.with(|java_activity| {
         *java_activity.borrow_mut() = Some(JavaActivity::new(env.clone(), activity));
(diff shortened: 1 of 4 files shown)


gl/src/lib.rs
```diff
@@ -607,6 +619,7 @@ impl Device for GLDevice {
         unsafe {
             gl::BindFramebuffer(gl::FRAMEBUFFER, framebuffer.gl_framebuffer); ck();
             gl::Viewport(0, 0, framebuffer.texture.size.x(), framebuffer.texture.size.y()); ck();
+            gl::Scissor(0, 0, framebuffer.texture.size.x(), framebuffer.texture.size.y()); ck();
```
pcwalton: Why do you set the scissor box? (I don't think it does anything unless `GL_SCISSOR_TEST` is enabled anyway, no?)

asajeffrey:
Rebased.

asajeffrey:
Rebased.

demo/android/rust/src/lib.rs
```diff
@@ -167,6 +170,28 @@ impl Window for WindowImpl {
...
+    fn view_box_size(&self, mode: Mode) -> Point2DI32 {
+        let mut width = self.size.x();
+        let height = self.size.y();
+        if let Mode::VR = mode {
+            width = width / 2;
+        }
+        Point2DI32::new(width as i32, height as i32)
+    }
+
+    fn make_current(&mut self, mode: Mode, index: Option<u32>) -> RectI32 {
```
pcwalton: I think this is a bit confusing for programmers who are familiar with OpenGL. `glxMakeCurrent`, `eglMakeCurrent`, etc. just make the context current without setting the viewport. How about just having another method `viewport` that subsumes `view_box_size` and the size calculation of this method? That will also avoid a bit of code duplication (the halving of `width`).

demo/android/rust/src/lib.rs
```diff
@@ -167,6 +170,28 @@ impl Window for WindowImpl {
         Point2DI32::new(0, 0)
     }
 
+    fn view_box_size(&self, mode: Mode) -> Point2DI32 {
```
pcwalton: How about something like:

    fn viewport(&self, view: View) -> Rect2DI32 {
        match view {
            View::Mono => { ... }
            View::Stereo(eye) => { ... }
    }

See below for rationale.

asajeffrey:
Rebased and split out getting the viewport from `make_current`.

PR #117, opened 2019-03-21
A couple of features for Options

asajeffrey:
A couple of changes to option-handling:
- allow demo apps to provide default options
- add an option to disable the UI


asajeffrey:
Rebased.

asajeffrey:
Review ping?

pcwalton:
I think turning off the UI would be good in general. Maybe allow the Tab key to cycle through "full UI", "only debug window", "no UI" options, and then add a command line switch to start in one of those modes.

asajeffrey:
Done.

PR #118, opened 2019-03-22
Add black background, and a bg option

asajeffrey:
Allows a black background.


asajeffrey:
Rebased.

pcwalton:
I was giving this some thought, and I was thinking what might work best is to change the "sun/moon" option slider into a popup that brings up a menu with "Light", "Dark", and "Transparent" options. Transparent would be what AR uses. What do you think?

asajeffrey:
Fine by me.

pcwalton:
This looks good once conflicts are resolved.

asajeffrey:
A couple of thoughts...

If we have a transparent background, what should the `native` demo display? Black?

How about we put off the replacement of the buttons with a menu for a different PR?

pcwalton:
> If we have a transparent background, what should the native demo display? Black?

Well, *ideally* it would ask SDL to recreate the window in transparent mode. (We don't want to have SDL always create a transparent window because that will slow all rendering down in the opaque case by 4ms or more.)

That may be a lot of work, though. Black is OK for now (though honestly we shouldn't even display the "transparent" option in that case).

> How about we put off the replacement of the buttons with a menu for a different PR?

asajeffrey:
@pcwalton Care to press the big green merge button?

asajeffrey:
Rebased.

pcwalton:
Commit c5ccd0f6e02b7e06d71f141a4a674faf75e436e6 implements this.

PR #122, opened 2019-03-25
In a VR demo, run the eyes in parallel

asajeffrey:
Running the two eyes concurrently takes the CPU time down from ~1.8ms to ~1.4ms on my iMac.


asajeffrey:
Rebased. Review ping @pcwalton?

asajeffrey:
It looks like pipelining makes this optimization difficult, since all the messages for camera 0 have to arrive before any of the messages for camera 1, so we'd need buffering to run the two cameras concurrently.

pcwalton:
Also, pipelining already achieves some of the gains from this, because it runs the sequential bottleneck in parallel with the tiling.

pcwalton:
This is no longer relevant with reprojection.

PR #123, opened 2019-03-28
 Add option to disable pipelining 

asajeffrey:
We could do this or the more drastic version which is just to remove pipelining completely.


asajeffrey:
Will this PR be redundant once pf3-pipelining lands?

pcwalton:
Yes, it should be.

asajeffrey:
OK, we can close this PR in that case.

PR #124, opened 2019-03-28
Allow the Window to customize rayon's thread builder

asajeffrey:
This is useful for platforms like magicleap which need to set up CPU affinity for threads.

### DIFF
--- a/demo/common/src/lib.rs
+++ b/demo/common/src/lib.rs
@@ -120,8 +120,8 @@ impl<W> DemoApp<W> where W: Window {
 
         // Set up Rayon.
         let mut thread_pool_builder = ThreadPoolBuilder::new();
-        thread_pool_builder = options.customize_rayon(thread_pool_builder);
-        thread_pool_builder = window.customize_rayon(thread_pool_builder);
+        thread_pool_builder = options.adjust_thread_pool_settings(thread_pool_builder);
+        thread_pool_builder = window.adjust_thread_pool_settings(thread_pool_builder);
         thread_pool_builder.build_global().unwrap();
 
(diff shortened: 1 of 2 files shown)


demo/common/src/lib.rs
```diff
@@ -118,6 +118,12 @@ impl<W> DemoApp<W> where W: Window {
 
         let view_box_size = view_box_size(options.mode, &window_size, false);
 
+        // Set up Rayon.
+        let mut thread_pool_builder = ThreadPoolBuilder::new();
+        thread_pool_builder = options.customize_rayon(thread_pool_builder);
```
pcwalton: I think this might be simpler if it were something like:

    let mut thread_pool_builder = window.thread_pool_builder();
    options.adjust_thread_pool_settings(&mut thread_pool_builder);
    thread_pool_builder.build_global().unwrap();

What do you think?

asajeffrey: Getting the window to build the builder is to problem, but annoyingly the builder methods all take a `self` and return a `Self`, so the options can't really take a `&mut` builder. We could make it:
```rust
let mut thread_pool_builder = window.thread_pool_builder();
thread_pool_builder = options.adjust_thread_pool_settings(thread_pool_builder);
thread_pool_builder.build_global().unwrap();
```
but I'm not sure if that's an improvement over what's in the PR. I quite like the fact that `Options` and `Window` have the same API for configuring the thread pool.

pcwalton: OK, sure. How about renaming `customize_rayon` to `adjust_thread_pool_settings` for clarity? r+ with that.

PR #125, opened 2019-03-29
Get pathfinder with the pf-no-simd feature to build on stable

asajeffrey:
Switching off simd should also switch off needing the simd nightly features.


pcwalton:
This is already in the pf3 branch.

PR #129, opened 2019-04-11
Made BackgroundColor public

asajeffrey:
Trying to use `BackgroundColor` outside `pathfinder_demo` results in:

```
error[E0603]: enum `BackgroundColor` is private
  --> demo/magicleap/src/lib.rs:28:22
   |
28 | use pathfinder_demo::BackgroundColor;
   |                      ^^^^^^^^^^^^^^^
```
or
```
error[E0603]: module `ui` is private
  --> demo/magicleap/src/lib.rs:28:22
   |
28 | use pathfinder_demo::ui::BackgroundColor;
   |                      ^^
```

### DIFF
--- a/demo/common/src/lib.rs
+++ b/demo/common/src/lib.rs
@@ -11,7 +11,7 @@
 //! A demo app for Pathfinder.
 
 use crate::device::{GroundLineVertexArray, GroundProgram, GroundSolidVertexArray};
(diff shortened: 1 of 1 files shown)


pcwalton:
I have an informal policy of no `pub use`, because I think it makes crates harder to understand. I landed this without it in 782e352bf28e2c892415ce0601e420eb0b52eb81.

PR #131, opened 2019-04-11
[WIP] Recycle vectors in assembly thread

asajeffrey:
This change removes most of the calls into jemalloc from the assembly thread. Without it, we were spending ~30% of CPU time in the allocator.


asajeffrey:
This is still an experiment ATM, not for merging!

asajeffrey:
Other perf improvements have made this less vital. Might still be worth revisiting at some point.

PR #132, opened 2019-04-15
Allow pf3 demo to work slightly more smoothly on Windows.

bzm3r:
One must still do all the work required to set up SDL, but at least the compiler won't complain about jemalloc, and git won't ask you if you want to add whatever SDL dll you need to dump in the pathfinder folder?


pcwalton:
Some of this is done now.

Issue #135, opened 2019-04-20
Constrain sizes of solid tile and alpha tile batches being sent to the GPU


bzm3r:
Fill tile batches are constrained (max of 16384?) thanks to some recent work: https://github.com/pcwalton/pathfinder/commit/c688d04412ec4324c8a3b68bded1a89e0737c23e

Something similar should be done for solid and alpha tiles too.

This will help when using newer graphics APIs which require buffer sizes to be specified upon buffer creation.

PR #148, opened 2019-05-11
Constrain sizes of solid tile and alpha tile batches being sent to the GPU

bzm3r:
This is an attempt to address: https://github.com/pcwalton/pathfinder/issues/135

Tiger seems to render correctly.


pcwalton:
This looks good, thanks!

pcwalton:
I reverted this because it broke the tiger when zooming in. See:

Screen Shot 2019-05-13 at 6 21 55 PM

PR #152, opened 2019-05-16
Silence unused variable warning

luisbg:
Silencing the following rustc warning:
warning: unused variable: `other`

Small fix but I want to contribute more if you are OK with it 😄

PR #160, opened 2019-05-24
Remove cross method from F32x4

luisbg:
Silencing the following rustc warning:
warning: unused variable: `other`

Followup to Pull Request #152 
Instead of marking `other` unused (`_other`) we can remove the method
entirely.


pcwalton:
Thanks!

luisbg:
Thanks for the review and merging!

Issue #147, opened 2019-05-09
Write a Godot/Unity/etc. plugin


pcwalton:
Several people have expressed interest in an HTML canvas-like API for Unity and so forth, and Pathfinder would be a great solution for this.

toolness:
Okay, I think I am going to try giving this a shot (for Unity at least). Will report back in a few days!

pcwalton:
Awesome! Let me know if you have any questions!

pcwalton:
I think you will probably want to first make a C API binding for Pathfinder's canvas API.

toolness:
Ok, sounds good!

I am not super familiar with Pathfinder itself, so I am learning as I go.  After looking at [`examples/canvas_minimal `](https://github.com/pcwalton/pathfinder/blob/master/examples/canvas_minimal/src/main.rs) it looks like the `CanvasRenderingContext2D` is largely independent of the Pathfinder renderer; it seems like this is the only place any "hand-off" occurs between the canvas and the renderer:

https://github.com/pcwalton/pathfinder/blob/a5d373cb91a6f49a7f0019e1f4a4b9180d840bda/examples/canvas_minimal/src/main.rs#L77-L78

So I'm curious how this hand-off should work in a potential C API.  You mentioned [on Twitter](https://twitter.com/pcwalton/status/1126307967622496256) that I need to wire up `RenderCommand` to Unity’s `CommandBuffer` abstraction. and it looks like `RenderCommand` is defined here:

https://github.com/pcwalton/pathfinder/blob/a5d373cb91a6f49a7f0019e1f4a4b9180d840bda/renderer/src/gpu_data.rs#L29-L37

I'm also noticing that the [`SceneProxy`](https://github.com/pcwalton/pathfinder/blob/54672f20898ba6c105f49364e82790641991cb90/renderer/src/concurrent/scene_proxy.rs) created in the minimal canvas example has all kinds of methods to output `RenderCommand`s.

Would a C API then have some kind of function that serializes the state of a `CanvasRenderingContext2D` into `RenderCommand`s that client code gobbles up?  The Unity plugin would wire these up into Unity's `CommandBuffer` abstraction, while other clients would do other things with them?

I thought it might be neat to have a basic proof-of-concept C program that uses the C API to render a canvas drawing to a window via SDL2.  I thought doing that might be a nice stepping-stone to building the Unity plugin, but I'm not sure.

Anyways, let me know what you think!

toolness:
[...]

PR #167, opened 2019-05-27
Add GLDevice::set_default_framebuffer()

toolness:
Okay, I have no idea if this PR contains the right kind of change for pathfinder, but I figured it'd start a conversation, at least.

Basically, when addressing the problem of dynamically resizing the in-editor Unity player mentioned in https://github.com/pcwalton/pathfinder/issues/147#issuecomment-496015315 (e.g., when the user clicks "play" and then resizes their window), I found that the approach used in Pathfinder's demo code didn't fully work in Unity:

https://github.com/pcwalton/pathfinder/blob/bc713a5181ab7ed90469d054c36f78baed36b306/demo/common/src/lib.rs#L266-L274

Specifically, what happens is that the Pathfinder rendering simply disappears every time I resize the window, though it occasionally/randomly comes back into view.  I tried simply re-instantiating the whole `Renderer` instance every time the window is resized, and while it did fix the problem, it was also very slow and made resizing the Unity window quite sluggish.

After some diagnosis I noticed that Unity appears to change the current GL framebuffer when the user resizes the window; I think the reason the rendering randomly re-appeared on resize was because eventually Unity would happen to use the same framebuffer ID that I'd initialized the renderer with.

However, I also found that if I'm able to _change_ the `GLDevice` that Pathfinder is using to reflect the new framebuffer ID, everything seems to work fine.  There isn't currently a way to do that, however, so this PR adds a simple accessor method that makes it possible.  But I'm not sure if this violates encapsulation in some way, or has other consequences that I'm not aware of.

For reference, the experimental Unity Plugin code that uses this accessor method is at https://github.com/toolness/pathfinder-unity-fun/commit/39221cf7b9bf7d5aba77cc4f7dcae91586eb8040.

@pcwalton do you have thoughts on this?

### DIFF
--- a/gl/src/lib.rs
+++ b/gl/src/lib.rs
@@ -42,6 +42,10 @@ impl GLDevice {
         }
     }
 
+    pub fn set_default_framebuffer(&mut self, framebuffer: GLuint) {
+        self.default_framebuffer = framebuffer;
+    }
+
     fn set_texture_parameters(&self, texture: &GLTexture) {
         self.bind_texture(texture, 0);
         unsafe {


pcwalton:
Yeah, this is the right fix. Thanks!

PR #171, opened 2019-05-28
Add test_i32x4_swizzles

luisbg:
This raises the test coverage in SIMD from 10.04% to 10.27%, just a bit.

@pcwalton, I see there is a comment about completing I32x4 swizzles, do you want me to do that? In this commit I'm only testing the 4 that exist.


pcwalton:
@luisbg Sure, feel free to complete them if you'd like.

PR #172, opened 2019-05-28
Complete I32x4 swizzles

luisbg:
This PR both completes all swizzles for I32x4 and adds tests for them.

Test coverage is now 20.90% because all new lines are covered.

I will add more tests for existing code next.


pcwalton:
Thanks!

PR #173, opened 2019-05-29
Start adding tests for scalar::F32x4

luisbg:
This isn't complete but I wanted to check with you that the structure is OK.

### DIFF
--- a/simd/src/test.rs
+++ b/simd/src/test.rs
@@ -9,6 +9,7 @@
 // except according to those terms.
 
 use crate::default::{F32x4, I32x4, U32x4};
+use crate::scalar::F32x4 as F32x4S;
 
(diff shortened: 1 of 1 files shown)


simd/src/test.rs
```diff
@@ -9,6 +9,7 @@
 // except according to those terms.
 
 use crate::default::{F32x4, I32x4, U32x4};
+use crate::scalar::F32x4 as sF32x4;
```
pcwalton: nit: Rust conventions want capital letters for type names. How about `F32x4S`?

luisbg:
Updated.

I forgot to say this raises the coverage to 21.37%.

pcwalton:
Thanks!

PR #177, opened 2019-06-04
Update README.md

jrr:
Travis badge and a note about demos


pcwalton:
Thanks! I think eventually we'd like to expand that section into an overview of all the directories, to help folks navigate the project.

PR #181, opened 2019-06-05
full build in CI

jrr:
- add cmake via apt
- replace demo builds with root-level `cargo build`

`cargo test` works, too, but there are a couple failing tests so I commented it out.


pcwalton:
Thanks!

PR #182, opened 2019-06-05
Define ck() based on cfg(debug_assertions)

toolness:
While attempting to investigate https://github.com/servo/pathfinder/issues/178, I noticed that the `ck()` function, which checks to see if there's a GL error and panics if so, wasn't running on my debug build.  Its definition was conditional based on whether `cfg(debug)` is set, but I'm not actually sure what that is or how to set it, whereas [`cfg(debug_assertions)` does the trick](https://stackoverflow.com/a/39205417/2422398).

This PR changes the conditional compilation to be based on `debug_assertions`.  However, if that's not the right way to approach this, please let me know!

### DIFF
--- a/gl/src/lib.rs
+++ b/gl/src/lib.rs
@@ -918,7 +918,7 @@ impl GLVersion {
 
 // Error checking
 
-#[cfg(debug)]
+#[cfg(debug_assertions)]
 fn ck() {
     unsafe {
         let err = gl::GetError();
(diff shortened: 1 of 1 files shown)


toolness:
I dug into GL error handling a bit more and found out that there are [only a handful of standard GL errors](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glGetError.xhtml) so I made the panic message a bit more helpful in 7563e95.

pcwalton:
Yikes, good catch!

PR #190, opened 2019-06-07
Make the backgrounds of the demo SVGs off-white

### LINKED ISSUE #144: Change white backgrounds of demo SVGs to off-white
According to Magic Leap contacts, bright white stresses the optics of the headset in ways that reduce performance. Changing any demos that involve bright white to an off-white colour instead should make the demo even better.

asajeffrey:
Fixes #144.


pcwalton:
I think I'd prefer to do this by simply having a filter that adjusts the colors in the color palette at runtime. That would be more reliable with custom SVGs.

asajeffrey:
I ended up picking different colours for each SVG, and leaving in a little bit of white to make details pop (e.g. the eyes of the magicleap logo). So I think the result is nicer than we'd get programmatically.

pcwalton:
Fair enough.

PR #193, opened 2019-06-08
Assert vertex attr locations exist in GLDevice.

toolness:
As @benpye discovered in https://github.com/servo/pathfinder/issues/183#issuecomment-500097365, Pathfinder crashes on systems with Nvidia drivers at least in part because it's looking up locations for vertex attributes that don't exist in some shaders.  In the spirit of "crash early, crash often", this PR adds assertions that ensure the vertex attribute ~~and uniform locations~~ we return are valid in `GLDevice`.  This doesn't fix any problems, of course, it just makes it easier to be aware of their existence on non-Nvidia systems, and makes them easier to debug.

That said, I'm not sure if this actually the best solution: perhaps `Device::get_vertex_attr()` and `Device::get_uniform()` should actually return `Option`-wrapped values, so that code calling it can change its behavior based on whether the attributes/uniforms exist or not?  I guess I'm not familiar enough with Pathfinder's codebase to know the best answer.  Regardless, no worries if you decide not to merge this.

### DIFF
--- a/gl/src/lib.rs
+++ b/gl/src/lib.rs
@@ -305,18 +305,24 @@ impl Device for GLDevice {
     }
 
     fn get_vertex_attr(&self, program: &Self::Program, name: &str) -> GLVertexAttr {
-        let name = CString::new(format!("a{}", name)).unwrap();
-        let attr = unsafe {
-            gl::GetAttribLocation(program.gl_program, name.as_ptr() as *const GLchar) as GLuint
+        let cname = CString::new(format!("a{}", name)).unwrap();
+        let attr_int = unsafe {
+            gl::GetAttribLocation(program.gl_program, cname.as_ptr() as *const GLchar)
         }; ck();
-        GLVertexAttr { attr }
+        if attr_int == -1 {
+            panic!("Vertex attribute '{}' does not exist in program", name);
+        }
+        GLVertexAttr { attr: attr_int as GLuint }
     }
 
     fn get_uniform(&self, program: &GLProgram, name: &str) -> GLUniform {
-        let name = CString::new(format!("u{}", name)).unwrap();
+        let cname = CString::new(format!("u{}", name)).unwrap();
         let location = unsafe {
-            gl::GetUniformLocation(program.gl_program, name.as_ptr() as *const GLchar)
+            gl::GetUniformLocation(program.gl_program, cname.as_ptr() as *const GLchar)
         }; ck();
+        if location == -1 {
+            panic!("Uniform '{}' does not exist in program", name);
+        }
         GLUniform { location }
     }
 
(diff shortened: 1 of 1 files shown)


gl/src/lib.rs
```diff
@@ -305,18 +305,24 @@ impl Device for GLDevice {
...
     fn get_vertex_attr(&self, program: &Self::Program, name: &str) -> GLVertexAttr {
-        let name = CString::new(format!("a{}", name)).unwrap();
-        let attr = unsafe {
-            gl::GetAttribLocation(program.gl_program, name.as_ptr() as *const GLchar) as GLuint
+        let cname = CString::new(format!("a{}", name)).unwrap();
+        let attr_int = unsafe {
+            gl::GetAttribLocation(program.gl_program, cname.as_ptr() as *const GLchar)
         }; ck();
-        GLVertexAttr { attr }
+        if attr_int == -1 {
```
toolness: We could wrap this `if` (and the corresponding one in `get_uniform`) in a `#[cfg(debug_assertions)]` if we're concerned about the performance.  I figured maybe it'd be OK not to do so since presumably these locations don't have to be looked up too often, but I dunno.

toolness:
One thing I noticed while running `canvas_minimal` with these assertions enabled is that GLSL uniforms which are declared in code but _unused_--and therefore optimized away during compilation--will trigger the assertion.

Currently this is happening when `SolidTileMulticolorProgram` is trying to retrieve the location of the `PaintTextureSize` uniform from `tile_solid_multicolor.vs.glsl`.  It's _defined_, it's just optimized out because it's not actually used (well, at least with Nvidia drivers, perhaps others will keep them around?).

gl/src/lib.rs
```diff
@@ -305,18 +305,24 @@ impl Device for GLDevice {
...
+            panic!("Vertex attribute '{}' does not exist in program", name);
```
toolness: Oops, I didn't realize that we were actually munging the names before passing them to OpenGL... I'm not sure if it'd be more helpful to panic with the name of the munged or un-munged name here.

pcwalton:
So we probably shouldn't be checking uniforms, because -1 is a valid uniform location (though it does nothing). However, for attributes we certainly should.

I think these checks should be debug mode only `debug_assert!()`, etc.

toolness:
Oh wow, I didn't realize that -1 is a valid uniform location.  Cool, I've just modified the PR to only pay attention to vertex attributes, and use `debug_assert!()`!

pcwalton:
So it turns out this change came for free with some of the Metal work I've been doing, and I cherry-picked it over to master here: https://github.com/servo/pathfinder/commit/dbf02fb742c6b673bb2f48188fbd054517dc02ed

This PR is obsolete now, but thanks for letting me know about the issue and putting it together :)

toolness:
Oh cool, sounds good!

Issue #178, opened 2019-06-04
Moving away from triangle fans makes the Unity plugin crash


toolness:
Ahoy! I was just updating the Unity plugin to the latest rev of Pathfinder in https://github.com/toolness/pathfinder-unity-fun/pull/2 and noticed that, as of acf666b70161228f139b8cc9c4e03a75a5c1101c, "Stop using triangle fans for DX12/Vulkan compatibility", the plugin can't render more than one frame before segfaulting.  Using the revision before it works fine, however.

Any idea what might be going on here, @pcwalton?

pcwalton:
Not without more information. Do you have a stack trace?

toolness:
Oops, sorry about that!  This is the exception:

```
Exception thrown at 0x00000000765479E0 (nvoglv64.dll) in unity-project.exe: 0xC0000005: Access violation reading location 0x0000000000000000. occurred
```

Here's the stack trace:

```
 	nvoglv64.dll!00000000765479e0()	Unknown
 	nvoglv64.dll!0000000075d549d7()	Unknown
>	GfxPluginPathfinder.dll!gl::DrawElementsInstanced(unsigned int mode, int count, unsigned int type_, core::ffi::c_void * indices, int instancecount) Line 1842	Unknown
 	GfxPluginPathfinder.dll!pathfinder_gl::{{impl}}::draw_elements_instanced(pathfinder_gl::GLDevice * self, pathfinder_gpu::Primitive primitive, unsigned int index_count, unsigned int instance_count, pathfinder_gpu::RenderState * render_state) Line 564	Unknown
 	GfxPluginPathfinder.dll!pathfinder_renderer::gpu::renderer::Renderer<pathfinder_gl::GLDevice>::draw_alpha_tiles<pathfinder_gl::GLDevice>(unsigned int self) Line 552	Unknown
 	GfxPluginPathfinder.dll!pathfinder_renderer::gpu::renderer::Renderer<pathfinder_gl::GLDevice>::render_command<pathfinder_gl::GLDevice>(pathfinder_renderer::gpu_data::RenderCommand * self) Line 251	Unknown
 	GfxPluginPathfinder.dll!pathfinder_renderer::concurrent::scene_proxy::SceneProxy::build_and_render<pathfinder_gl::GLDevice>(pathfinder_renderer::gpu::renderer::Renderer<pathfinder_gl::GLDevice> * self, pathfinder_renderer::options::RenderOptions renderer) Line 88	Unknown
[...]

PR #197, opened 2019-06-15
Fix Unity plugin crash and add debug assertions

toolness:
Okay, I did some investigating into #178 and discovered a few things.

Right now, the Unity plugin instantiates a `Renderer` only once and reuses it every frame (instantiating one every frame seemed excessive, but I can alter this behavior if you want).  However, as of acf666b70161228f139b8cc9c4e03a75a5c1101c, the `quad_vertex_indices_buffer` is only bound to `GL_ELEMENT_ARRAY_BUFFER` when the renderer is instantiated--not every frame.

This seems fine if we assume that other GL code hasn't changed the binding since the renderer was instantiated.  I _think_ this is what's going on with the Unity plugin: because the plugin has to share the same GL context with Unity's own renderer, it's likely that Unity is changing the `GL_ELEMENT_ARRAY_BUFFER` binding when it's doing its own drawing.

Furthermore, the stack trace mentioned at https://github.com/servo/pathfinder/issues/178#issuecomment-498854709 is happening because [the `indices` parameter of `glDrawElementsInstanced()` is defined in a confusing way](https://stackoverflow.com/questions/21706113/the-4th-argument-in-gldrawelements-is-what): if `GL_ELEMENT_ARRAY_BUFFER` is bound, then it's interpreted as an offset index into an array (which is how we're using it), but if it's _not_ bound anymore--as is the case when running in Unity--it's interpreted as a pointer value, in which case the program will segfault.

I'm not really sure what the best solution is here.  I've added a debug assertion called `assert_element_array_buffer_is_bound()` to make us panic with a helpful error instead of causing a null pointer dereference, which will hopefully help us catch errors like this in the future, but I'm not sure what the best strategy to ensure the proper buffer binding is.  If we only plan to ever bind `quad_vertex_positions_buffer` to `GL_ELEMENT_ARRAY_BUFFER` then we could just bind it once during `Renderer::begin_scene()`, but otherwise we may need to bind it immediately before we call `glDrawElementsInstanced()` or anything else that needs it.

For now I've just done the minimal possible required to make the minimal canvas example work in the Unity plugin, but it's possible the ideal solution is something different.  I'm open to suggestions, this PR is really just my way of starting the conversation.

### DIFF
--- a/gl/src/lib.rs
+++ b/gl/src/lib.rs
@@ -539,6 +539,7 @@ impl Device for GLDevice {
 
     fn draw_elements(&self, primitive: Primitive, index_count: u32, render_state: &RenderState) {
         self.set_render_state(render_state);
+        assert_element_array_buffer_is_bound();
         unsafe {
             gl::DrawElements(primitive.to_gl_primitive(),
                              index_count as GLsizei,
@@ -554,6 +555,7 @@ impl Device for GLDevice {
                                instance_count: u32,
                                render_state: &RenderState) {
         self.set_render_state(render_state);
+        assert_element_array_buffer_is_bound();
         unsafe {
             gl::DrawElementsInstanced(primitive.to_gl_primitive(),
                                       index_count as GLsizei,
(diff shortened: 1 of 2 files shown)


pcwalton:
Cool, thanks for digging in. My Metal changes in my `metal` branch will hopefully fix this, since they rework the API around VBO/IBO binding, so let's try again once those have landed.

toolness:
Oh cool, sounds good!

pcwalton:
@toolness I just pushed the Metal changes. Want to try again?

toolness:
Yes, this works perfectly and logs no weird GL warnings either!! Thanks!

PR #216, opened 2019-07-14
Expose Canvas transform and save/restore to C

toolness:
This exposes the Canvas' transform and save/restore functionality to the C API.

The only method I haven't added here is `Canvas::current_transform()`, which I can add for completeness if you want.

### DIFF
--- a/c/src/lib.rs
+++ b/c/src/lib.rs
@@ -321,6 +321,16 @@ pub unsafe extern "C" fn PFCanvasResetTransform(canvas: PFCanvasRef) {
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn PFCanvasSave(canvas: PFCanvasRef) {
+    (*canvas).save();
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn PFCanvasRestore(canvas: PFCanvasRef) {
+    (*canvas).restore();
+}
+
+#[no_mangle]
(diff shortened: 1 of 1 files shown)


c/src/lib.rs
```diff
@@ -310,6 +310,26 @@ pub unsafe extern "C" fn PFCanvasSetLineDash(canvas: PFCanvasRef,
     (*canvas).set_line_dash(slice::from_raw_parts(new_line_dashes, new_line_dash_count).to_vec())
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn PFCanvasSetCurrentTransform(canvas: PFCanvasRef, transform: *const PFTransform2F) {
+    (*canvas).set_current_transform(&(*transform).to_rust());
```
toolness: This is interesting because it seems like we're ultimately copying the transform twice--once to make it owned by Pathfinder, but then the canvas copies it again to make it its own.  Is there a way to optimize this to only be one copy?  Or will Rust actually be super smart and optimize things at compile time?

toolness:
Hmm, I am not sure what is going on with Travis--looks like cargo is failing with `error trying to exec 'cc1obj': execvp: No such file or directory` and that doesn't seem to be related to the content of my PR.

c/src/lib.rs
pcwalton: It shouldn't be a big deal either way. None of the canvas methods are optimized that well to begin with—the performance-critical parts are the scene building.

Issue #202, opened 2019-06-22
Raw glyph drawing function for the canvas API


pcwalton:
This is useful in some cases.

s3bk:
This is what I added to draw glyphs.
```rust
    pub fn fill_layout(&mut self, layout: &Layout, transform: Transform2DF) {
        let paint_id = self.scene.push_paint(&self.current_state.fill_paint);
        drop(self.scene.push_layout(&layout,
                                    &TextStyle { size: self.current_state.font_size },
                                    &transform.post_mul(&self.current_state.transform),
                                    TextRenderMode::Fill,
                                    HintingOptions::None,
                                    paint_id));
    }
```

I am using `skribo::Layout here` as that is what is used by `push_layout`.
The transformation matrix is used instead of a position, because PDF has two different coordinate spaces for graphics (positive y -> down) and text (positive y -> up)…

PR #218, opened 2019-07-16
fixes and a possible canvas addition

s3bk:
What I am not sure about is `fill_layout` (see #202). I added it because it was needed to push layouts directly, but currently I am working on removing the dependency on text altogether. (And instead push the glyph outlines).


swf/src/shapes.rs
```diff
@@ -93,12 +93,12 @@ impl Shape {
     }
 
     #[inline]
-    fn first(&self) -> LineSegment {
+    fn first(&self) -> &LineSegment {
```
pcwalton: Why this change? `LineSegment` fits in a single SIMD register; there's little point in returning references.

pcwalton:
`fill_layout` is perfectly fine. I was thinking of adding it myself actually.

swf/src/shapes.rs
s3bk: There was an error and I simply applied the suggested fix. No other reason.

s3bk: Dereferenced it now instead.

swf/src/shapes.rs
```diff
@@ -94,12 +94,12 @@ impl Shape {
 
     #[inline]
     fn first(&self) -> LineSegment {
-        self.outline.first().unwrap()
+        *self.outline.first().unwrap()
```
pcwalton: Will this still compile?

s3bk: It compiles for me (`cargo build -p pathfinder_swf`)

s3bk: without:
```
error[E0308]: mismatched types
   --> swf/src/shapes.rs:102:9
    |
101 |     fn last(&self) -> LineSegment {
    |                       ----------- expected `shapes::LineSegment` because of return type
102 |         self.outline.last().unwrap()
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |         |
    |         expected struct `shapes::LineSegment`, found reference
    |         help: consider dereferencing the borrow: `*self.outline.last().unwrap()`
    |
    = note: expected type `shapes::LineSegment`
               found type `&shapes::LineSegment`
```

pcwalton:
Hmm, we shouldn't return a reference from `Outline::last()` then, but let's fix that in a follow up.

s3bk:
```
    pub(crate) outline: Vec<LineSegment>, // Could be Vec<(start, end)>
```
not much that can be done here …

PR #225, opened 2019-07-27
Make text rendering optional in pathfinder_canvas

### LINKED ISSUE #164: Make skribo/HarfBuzz use optional
This makes me a bit sad, but given that binary size is important for wasm especially, and since there are some use cases that don't need text at all (or if they do need it it's only for debugging), we should probably have an optional dumb Latin-only shaper.

Also, HarfBuzz is our only C dependency, and it would be nice to have some kind of Rust-only version, in case a C compiler isn't handy.

### LINKED ISSUE #208: Make FreeType optional
And with that Font-Kit.
It is a [blocker for WASM](https://github.com/servo/pathfinder/issues/155) and not everyone needs it anyway.

(The [PDF viewer](https://github.com/s3bk/pdf/blob/master/view/src/lib.rs) currently still uses FreeType, but I hope to switch it to [RustType](https://gitlab.redox-os.org/redox-os/rusttype) soon.

est31:
Fixes #164
Fixes #208

### DIFF
--- a/c/Cargo.toml
+++ b/c/Cargo.toml
@@ -15,6 +15,7 @@ gl = "0.6"
 libc = "0.2"
 
 [dependencies.pathfinder_canvas]
+features = ["text"]
 path = "../canvas"
 
 [dependencies.pathfinder_content]
--- a/c/Cargo.toml
+++ b/c/Cargo.toml
@@ -15,7 +15,7 @@ gl = "0.6"
 libc = "0.2"
 
 [dependencies.pathfinder_canvas]
-features = ["text"]
+features = ["pf-text"]
 path = "../canvas"
 
 [dependencies.pathfinder_content]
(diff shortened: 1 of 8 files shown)


est31:
cc @sebk this is what you wanted I think

est31:
I've tested @s3bk 's pdf viewer with this patch and it works great.

s3bk:
thanks!
Do you think it would be cleaner to move the text stuff into canvas/src/text.rs ?

est31:
@s3bk yeah it's cleaner. I've pushed a commit to move the text stuff into separate files.

pcwalton:
I generally prefix all Pathfinder features with `pf-` to avoid conflicts with other crates.

est31:
re-r? @pcwalton

pcwalton:
Cool, thanks!

PR #235, opened 2019-08-22
WIP: iOS support

mtak-:
This requires a couple PRs to `usvg` and `harfbuzz-sys`.

The simd feature also appeared to be broken on iOS, but I did not spend much time looking into that.


mtak-:
If there's interest in having iOS support now on pathfinder, I'm happy to finish this PR. I got a hacked up version of the `native` demo working on iOS (and the tiger looks very pretty on the ipad pro). My company is switching priorities a bit, and plan on coming back to this later.

lkzhao:
@mtak- Any chance to bring this support back to life? Happy to sponsor as much as I can.

PR #243, opened 2019-11-30
Add a Glow device abstraction

jnetterf:
This implements GLOW (GL on Whatever) implementation of the device
abstraction. It is a very close copy of the GL implementation.

This abstraction can be used in multiple environments (Open GL, OpenGL
ES, and WebGL) and avoids target-specific code. The abstraction can be
used with wasm32-unknown-unknown + web-sys, stdweb, and natively.
I wrote this to enable limited WebGL 2 + wasm32-unknown-unknown support.

This also makes minor changes to shaders so that they work in WebGL 2
and implements an example that works both in web browsers and natively,
via SDL.


glow/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "pathfinder_glow"
+version = "0.1.0"
+edition = "2018"
+authors = ["Patrick Walton <pcwalton@mimiga.net>", "Joshua Netterfield <joshua@nettek.ca"]
```
pcwalton: nit: missing `>` at end of email

examples/canvas_glow/src/static_resource_loader.rs
```diff
@@ -0,0 +1,107 @@
+use pathfinder_gpu::resources::ResourceLoader;
+
+/// A hacky loader that includes the resources we need in the binary.
+///
+/// On desktop, we can use FilesystemResourceLoader (see canvas_minimal), but
+/// wasm32-unknown-unknown does not have a filesystem.
+///
+/// This does not have any platform dependencies, but will result in larger binaries, and slower
+/// compile times, so you may wish to implement an approach with different trade-offs.
```
pcwalton: Hmm, have you considered using XHR or the Fetch API instead?

pcwalton:
Also, we should just remove `pathfinder_gl` if we decide to go with GLOW. No sense having multiple versions of the GL backend.

examples/canvas_glow/src/static_resource_loader.rs
jnetterf: Yes. This is likely something you would want to do outside of a minimal example. I'll explicitly mention this option in the comment. It may also be nice to provide such a loader in the future.

There are two reasons we may wish to avoid using a web API in this example. First, it's target-specific code that is not core to the example. Secondly, it requires a fair bit of setup, making the example harder to follow along with.

What do you think?

pcwalton: Well, my main concern is that it hardcodes all the pathnames of resources, which adds some maintenance burden.

jnetterf:
Closing in favour of https://github.com/s3bk/pathfinder

PR #252, opened 2020-01-23
fix simd for scalar platforms

s3bk:


s3bk:
whoops. why is the webgl stuff in here…

s3bk:
appears to be fixed now.

pcwalton:
Thanks!

PR #254, opened 2020-01-28
Remove log restriction

arturoc:
pathfinder-geometry enabled a log feature that restricts logging to warn for any binary using it.

This is really hard to find for anyone using the package as suddenly logging seems to stop working for no apparent reason

### DIFF
--- a/demo/common/Cargo.toml
+++ b/demo/common/Cargo.toml
@@ -20,7 +20,6 @@ features = ["png_codec"]
 
 [dependencies.log]
 version = "0.4"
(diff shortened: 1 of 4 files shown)


s3bk:
This is actually enabled in a few more places. A `rg` will show.

Thanks for finding it! I already replaced the log crate because I could not find why log didn't work.

arturoc:
yes i've removed all places where it appears now

pcwalton:
I actually don't know why this got in there. Maybe it was added in a PR? Anyway, thanks for finding it!

PR #256, opened 2020-01-31
Add WebGL impl and enable access to Outline, Contour and Scene

s3bk:
This is a squashed commit of my cleaned up collected modifications.


shaders/tile_alpha_vertex.inc.glsl
```diff
@@ -33,7 +33,7 @@ void computeVaryings() {
     vec2 origin = vec2(aTileOrigin.xy) + vec2(aTileOrigin.z & 15u, aTileOrigin.z >> 4u) * 256.0;
     vec2 position = (origin + vec2(aTessCoord)) * uTileSize;
     vec2 maskTexCoordOrigin = computeTileOffset(uint(aTileIndex), uStencilTextureSize.x);
-    vec2 maskTexCoord = maskTexCoordOrigin + aTessCoord * uTileSize;
+    vec2 maskTexCoord = maskTexCoordOrigin + vec2(aTessCoord) * uTileSize;
```
s3bk: This turned out to make WebGL very unhappy

gpu/src/resources.rs
```diff
@@ -56,14 +57,26 @@ impl FilesystemResourceLoader {
...
             .for_each(|segment| path.push(segment));
 
         let mut data = vec![];
         File::open(&path)?.read_to_end(&mut data)?;
-        Ok(data)
+        Ok(data.into())
+    }
+}
+
+pub struct EmbeddedResourceLoader;
```
s3bk: This can probably use some tuning in the future to only include the files actually needed. But it works for now.

renderer/src/builder.rs
```diff
@@ -53,6 +58,7 @@ impl<'a> SceneBuilder<'a> {
     }
 
     pub fn build<E>(&mut self, executor: &E) where E: Executor {
+        #[cfg(not(target_arch = "wasm32"))]
         let start_time = Instant::now();
```
s3bk: caused a panic in wasm

renderer/src/options.rs
```diff
@@ -17,6 +17,7 @@ use pathfinder_geometry::transform3d::Perspective;
 use pathfinder_geometry::vector::{Vector2F, Vector4F};
 use pathfinder_content::clip::PolygonClipper3D;
 
+// FIXME: Can this use &mut self ? (Sebastian Köln)
```
s3bk: The idea is to make it a bit faster…

canvas/src/lib.rs
```diff
@@ -403,6 +403,29 @@ impl Path2D {
         }
     }
 
+    /// Add other path to this path.
```
s3bk: These functions are used by [vector](https://github.com/s3bk/vector)

canvas/src/lib.rs
```diff
@@ -403,6 +403,29 @@ impl Path2D {
...
+    
+    /// Transform the path with the given transformation.
+    /// returns the transformed path.
+    ///
+    /// note: any non-closed subpaths will be closed
+    pub fn transform(mut self, transform: &Transform2F) -> Path2D {
+        self.flush_current_contour();
+        self.outline.transform(transform);
+        self
+    }
```
s3bk: These functions are used by [vector](https://github.com/s3bk/vector)

pcwalton:
Hmm. Maybe we should move the WebGL code into `pathfinder_gl`, so that we can avoid duplicating helper functions like `flip_y`? We can use `#[cfg]` directives to choose the right implementation.

gpu/src/resources.rs
pcwalton: Won't this cause a rebuild when changing the shaders? One of the benefits of the `FilesystemResourceLoader` is live shader editing without rebuilding. I know it's not possible to use the filesystem in WebAssembly, but maybe we should make this enabled on the wasm target only.

s3bk:
flip_y doesn't actually appear to be used

s3bk:
hm. actually that broke something.

s3bk:
So the EmbeddedResourceLoader is broken and I am not sure why. I didn't change anything from when it was working…

s3bk:
working again.

gpu/src/resources.rs
s3bk: I moved the `EmbeddedResourceLoader` into pathfinder_resources and added a bunch of feature flags to control what gets included.
The demo runs.

s3bk:
As a side effect of adding fine control over what gets embedded, my wasm is now less than half the size.

s3bk:
The latest commit is not actually WebGL specific. However it fixed glitches that appear to originate from the tile background texture lookup.

Dollab:
This is really cool ! Would be awesome to merge this !

pcwalton:
Reviewing this now.

pcwalton:
WebGL backend is merged: https://github.com/servo/pathfinder/commit/e2fbde820d62b8ae74e4665e7f19036d55d1fb9b

pcwalton:
Made the WebGL fixes, I believe: https://github.com/servo/pathfinder/commit/ee6c7e07973b3ccce48be837dbe53eff8b9d5d74

zimond:
@pcwalton seems `pathfinder_resources` is not declared as a dep of `pathfinder_webgl`

PR #257, opened 2020-02-04
Bump almost all dependencies to their latest version

linkmauve:
I’m pleasantly surprised so few of them required any kind of (visible) code change. The few which did have a dedicated commit.

The only remaining one according to `cargo-outdated` is `pathfinder_text`’s dependency on `lyon_path`, but this is just waiting for a `font-kit` release as that crate dropped lyon from its direct dependencies anyway.

A `cargo test` goes down from 373 crates to 365 crates to build.


linkmauve:
Now that #259 has been merged, tests should pass and this should be good to go!

PR #258, opened 2020-02-05
Add a minimal example in C

linkmauve:
This is a direct port of `examples/canvas_minimal`, using GLFW instead of SDL2 and written in C to show how to use this binding.

### DIFF
--- a/examples/glfw_app/meson.build
+++ b/examples/glfw_app/meson.build
@@ -1,0 +1,15 @@
+project(
+  'pathfinder-example',
+  'c',
+  version: '0.1.0'
+)
+
+dep_glfw = dependency('glfw3')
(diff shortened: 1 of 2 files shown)


pcwalton:
Thanks! How about putting this in `examples/`, to go alongside the existing `macos_app` C example?

linkmauve:
I moved it to `examples/glfw_app/` to match the macOS example.

I’m planning on moving `pathfinder_c` to use lu_zero’s [`cargo-c`](https://github.com/lu-zero/cargo-c) instead of `bindgen` directly, this will give the benefit of creating a .pc file that can then be installed, and let other software use Pathfinder’s C API as they would with any other C library, with `pkg-config`.

PR #259, opened 2020-02-05
Add quickcheck build-dependency

linkmauve:
`quickcheck` usage in `pathfinder_content` had been added in b8f622203ae96776fe69a0c10282792832db9ef2, but not to the `Cargo.toml`.

This makes tests pass again.


pcwalton:
Thanks!

PR #277, opened 2020-03-01
Bump almost all dependencies to their latest version

linkmauve:
As in #257, I still left out the swf and lyon crates.

I also removed the default features of the `image` crate in area-lut and gamma-lut, to match how it is used in the rest of Pathfinder.


pcwalton:
Last time I merged a PR like this it broke Metal because the Metal crate relies on `foreign-types` 0.3.

linkmauve:
Oh, I didn’t realise because I don’t have access to any macOS computer.

I’ve downgraded `foreign-types` to 0.3 to fix it, but it’d be nice to have a way to test that from the CI.

c/src/lib.rs
```diff
@@ -11,7 +11,6 @@
 //! C bindings to Pathfinder.
 
 use font_kit::handle::Handle;
-use foreign_types::ForeignTypeRef;
```
pcwalton: Why remove this?

linkmauve: Because it wasn’t being used anywhere and gave me a warning.

Is it used only on macOS maybe?  If so, I will put it back under a `#[cfg]` conditional.

linkmauve: Here, I just readded it.

linkmauve:
This should now be good to go. :)

pcwalton:
This is merged as https://github.com/servo/pathfinder/commit/9723fa441b5a098aac9d3f37aac68d08669a0a4c, minus the part that broke Metal (the cocoa upgrade is incompatible).

PR #282, opened 2020-03-06
Add tests and debug impl for Path2D

inferiorhumanorgans:


canvas/src/lib.rs
```diff
@@ -610,6 +610,49 @@ impl Path2D {
...
+fn test_even_longer_closed_path() {
+    let mut path = Path2D::new();
+    path.move_to(Vector2F::new(0., 1.));
+    path.line_to(Vector2F::new(2., 3.));
+    path.line_to(Vector2F::new(4., 5.));
+    path.close_path();
+
+    let path_string = format!("{:?}", path.into_outline());
+    assert_eq!("M 0 1 L 2 3 L 4 5 L 0 1 z", path_string);
+}
```
pcwalton: Let's move these into a `tests.rs` file to avoid `lib.rs` getting too big.

PR #289, opened 2020-03-26
not actually a PR

s3bk:
I am making this to discuss changes I made and whether to merge some of them.


canvas/src/lib.rs
```diff
@@ -569,6 +569,29 @@ impl Path2D {
...
+    
+    /// Transform the path with the given transformation.
+    /// returns the transformed path.
+    ///
+    /// note: any non-closed subpaths will be closed
+    pub fn transform(mut self, transform: &Transform2F) -> Path2D {
+        self.flush_current_contour();
+        self.outline.transform(transform);
+        self
+    }
```
s3bk: These two can probably be added without problems. They are used by [vector](https://github.com/s3bk/vector).

content/src/outline.rs
```diff
@@ -120,6 +120,17 @@ impl Outline {
...
+    pub fn clear(&mut self) {
+        self.contours.clear();
+        self.bounds = RectF::default();
+    }
+    
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.contours.len()
+    }
+
```
s3bk: also used by vector

content/src/outline.rs
```diff
@@ -227,6 +238,16 @@ impl Contour {
...
 
+    #[inline]
+    pub fn with_capacity(length: usize) -> Contour {
+        Contour {
+            points: Vec::with_capacity(length),
+            flags: Vec::with_capacity(length),
+            bounds: RectF::default(),
+            closed: false,
+        }
+    }
```
s3bk: used by vector

content/src/outline.rs
```diff
@@ -308,6 +329,15 @@ impl Contour {
...
     }
 
+    #[inline]
+    pub fn clear(&mut self) {
+        self.points.clear();
+        self.flags.clear();
+        self.bounds = RectF::default();
+        self.closed = false;
+    }
+
```
s3bk: used by vector

log/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+macro_rules! pa_warn {
+    ($($t:tt)*) => ()
+}
+
+#[macro_export]
+macro_rules! pa_error {
+    ($($t:tt)*) => (
+        $crate::log::error!($($t)*)
+    )
+}
```
s3bk: I am not sure about this. But I also could not work without it.

renderer/src/builder.rs
```diff
@@ -84,12 +84,13 @@ pub(crate) struct Occluder {
...
-impl<'a> SceneBuilder<'a> {
+impl<'a, L: RenderCommandListener> SceneBuilder<'a, L> {
     pub(crate) fn new(
         scene: &'a Scene,
         built_options: &'a PreparedBuildOptions,
-        listener: Box<dyn RenderCommandListener>,
-    ) -> SceneBuilder<'a> {
+        listener: L,
+    ) -> SceneBuilder<'a, L> {
+        let effective_view_box = scene.effective_view_box(built_options);
```
s3bk: I replaced the trait object with a generic, as in WebAssembly it is all single threaded and are no runtime decisions possible. I did not measure the performance gain.

simd/src/scalar/mod.rs
```diff
@@ -507,6 +499,37 @@ impl I32x2 {
...
+        ])
+    }
+
+    #[inline]
+    pub fn packed_lt(self, other: I32x2) -> U32x2 {
+        U32x2([
+            if self[0] < other[0] { !0 } else { 0 },
+            if self[1] < other[1] { !0 } else { 0 },
+        ])
+    }
```
s3bk: These should probably be added.

content/src/outline.rs
s3bk: these are used to improve performance (and they do a good job at it).
Turns out copying the known length into a newly allocated buffer, and then reusing the old one for the next path, is a lot faster than starting with an empty buffer each time and transferring ownership.

renderer/src/builder.rs
pcwalton: I used a trait object to improve compilation time and because the overhead of vtable dispatch should be minimal.

log/src/lib.rs
pcwalton: Just because of compilation time?

s3bk: It can make things more difficult if you forget to enable the appropriate feature.
(Then again, most users would probably not have to debug pathfinder anyway.)
I ran into this myself…

s3bk: I just don't have any better solution…

pcwalton: I'm not sure what you mean by forgetting to enable the appropriate feature, could you elaborate?

content/src/outline.rs
s3bk: this one?

PR #343, opened 2020-05-18
Replace cbindgen with cargo-c

linkmauve:
cargo-c was written for rav1e’s usecase of generating and installing a C ABI the proper way, with a static library, a shared library, a pkg-config file, and the include files at the right location, and pathfinder has the exact same needs.

This requires the user to use a different command from the usual `cargo build`, which is why I added a README.md.

Ideally we’d add a path prefix for the include directory, to allow multiple versions to be installed at the same time until we freeze the API, but cargo-c doesn’t support it yet.


pcwalton:
This looks OK overall, except renaming the crate name. I filed https://github.com/lu-zero/cargo-c/issues/83.

PR #359, opened 2020-06-17
Various Outline things

s3bk:
Outline:
- `merge` combine two outlines
- `transformed` variant of `transform` that takes `self`
- `mirror_and_close` to mirror a Contour on the line between its first and last point
- `with_capacity'
- `len`
- `from_rect_rounded`

Contour:
 - `from_rect_rounded`: SVG style rect with rounded corners
 - `push_svg_arc`: SVG arc


pcwalton:
Could you elaborate as to what `mirror_and_close()` is for?

content/src/outline.rs
```diff
@@ -496,6 +566,64 @@ impl Contour {
         const EPSILON: f32 = 0.001;
     }
 
+    /// Push a SVG arc
+    ///
+    /// Draws an ellipse section with radii given in `radius` rotated by `x_axis_rotation` to 'to' in the given `direction`.
+    /// If `large_arc` is true, draws an arc bigger than a 180°, otherwise smaller than 180°.
+    /// note that `x_axis_rotation` is in radians.
+    pub fn push_svg_arc(&mut self, radius: Vector2F, x_axis_rotation: f32, large_arc: bool, direction: ArcDirection, to: Vector2F) {
```
pcwalton, line 586: How does this compare to the arc implementation that `pathfinder_canvas` uses?

content/src/outline.rs
```diff
@@ -293,6 +294,75 @@ impl Contour {
         contour
     }
 
+    #[inline]
+    pub fn from_rect_rounded(rect: RectF, radius: Vector2F) -> Contour {
+        use std::f32::consts::SQRT_2;
+        const QUARTER_ARC_CP_FROM_OUTSIDE: f32 = (3.0 - 4.0 * (SQRT_2 - 1.0)) / 3.0;
```
pcwalton, line 312: Where does this number come from?

s3bk: It is from https://www.chartwellyorke.com/mx/explore/geometric-modeling/circle-spline-approximation.pdf
page 4. Except that I am using 1 - 4/3 (√2-1) = 3/3 - 4/3 (√2-1) = (3 - 4(√2-1))/3

content/src/outline.rs
s3bk: Path2D::ellipse_section takes
- center
- axis
- start_angle
- end_angle: f32,
- direction

svg_arc takes 
 - last point (implicit)
 - radius
 - axis rotation
 - large arc flag, because there are two possible circles
 - destination point

s3bk:
> Could you elaborate as to what `mirror_and_close()` is for?

`mirror_and_close` isn't really necessary. It is a nice utility for constructing symmetric paths as it allows constructing only one half of it and then mirroring it automatically.

The symmetry line is between the first and last point on the existing Contour.

PR #384, opened 2020-07-04
PDF support for the demo

s3bk:


Cargo.toml
```diff
@@ -69,10 +69,3 @@ default-members = [
 [patch.crates-io]
 pathfinder_geometry = { path = "geometry" }
 pathfinder_simd = { path = "simd" }
-
-[patch."https://github.com/s3bk/pathfinder"]
-pathfinder_geometry = { path = "geometry" }
-pathfinder_simd = { path = "simd" }
-pathfinder_renderer = { path = "renderer" }
-pathfinder_content = { path = "content" }
-pathfinder_color = { path = "color" }
```
pcwalton: Please squash these commits into one

demo/common/src/lib.rs
```diff
@@ -81,14 +83,19 @@ mod device;
 mod renderer;
 mod ui;
 
+enum Content {
+    Svg(SvgTree),
+    Pdf(PdfFile<Vec<u8>>, PdfRenderCache, u32)
```
pcwalton, line 88: I think this should probably be a struct-like enum variant, so that we can give the `u32` field a name.

demo/common/src/lib.rs
```diff
@@ -430,25 +435,22 @@ impl<W> DemoApp<W> where W: Window {
                     }
                 }
 
-                Event::OpenSVG(ref svg_path) => {
+                Event::OpenData(ref data_path) => {
```
pcwalton, line 438: Maybe `OpenFile`?

demo/common/src/window.rs
```diff
@@ -114,7 +114,7 @@ pub struct OcularTransform {
 }
 
 #[derive(Clone)]
-pub enum SVGPath {
+pub enum DataPath {
```
pcwalton, line 117: Maybe `FilePath`?

demo/common/src/lib.rs
```diff
@@ -755,30 +757,45 @@ pub enum UIVisibility {
...
+        DataPath::Path(ref path) => std::fs::read(path).unwrap().into()
     };
 
-    let tree = Tree::from_data(&data, &UsvgOptions::default()).expect("Failed to parse the SVG!");
-    let built_svg = build_svg_tree(&tree, viewport_size, filter);
-    (built_svg, tree)
+    if let Ok(tree) = SvgTree::from_data(&data, &UsvgOptions::default()) {
+        Content::Svg(tree)
+    }
+    else if let Ok(pdf) = PdfFile::from_data(data) {
```
pcwalton, line 789: nit: weird indentation here

s3bk:
How about putting the PDF part behind a feature flag? It is a bit fragile…

demo/common/src/window.rs
s3bk: Technically, it doesn't have to be a file… I could see URLs useful too.
Maybe `DataSource`?

demo/common/src/lib.rs
s3bk: fixed

pcwalton:
It's just the demo. Lots of things in the demo are fragile. I'm not worried about it :)

PR #404, opened 2020-07-17
Make `canvas_minimal` example rerender on refresh

leroycep:
I'm not sure if this actually works, because the `canvas_minimal` example doesn't run on my computer.


pcwalton:
This breaks on macOS because `winit` doesn't send an initial `Refresh` event there.

leroycep:
Hmm, I can't find anything that looks like a "window shown" event. I think I'll move the rendering code again, and then use boolean to indicate that the screen needs to be redrawn.

Issue #422, opened 2020-07-28
cbindgen is not making use of cfg directives because of lack of [defines] section in cbindgen.toml 


BenGachowski:
The header file generated with cbindgen will automatically wrap the various exported functions in c preprocessor macros, but only if those macros are defined in the cbindgen.toml under a [defines] section.

For example, on my windows machine, the functions labeled for MacOS only are still included leading to undefined references when compiling.

More information about this can be [found here](https://github.com/eqrion/cbindgen/issues/105).

PR #424, opened 2020-07-29
Add [defines] section to cbindgen.toml

BenGachowski:
Attempts to fix [issue #422](https://github.com/servo/pathfinder/issues/422).


pcwalton:
Can we just write ` "target_os = macos" = "__APPLE__"`?

BenGachowski:
I don't see why not, that should work just fine.

PR #452, opened 2021-01-10
Fix arm simd shuffling arguments

### LINKED ISSUE #450: Invalid simd implementation on Apple M1
See https://github.com/servo/font-kit/issues/177 for details.

Building pathfinder_simd with feature "pf-no-simd" makes font-kit work again.

berkus:
According to packed_simd docs, the syntax for simd_shuffle4()
is as follows:

"The indices must be in range [0, M * N) where M is the number of input
vectors (1 or 2)and N is the number of lanes of the input vectors. The
indices i in range [0, N) refer to the i-th element of vec0, while the
indices in range [N, 2*N) refer to the i - N-th element of vec1."

I did not find implementation or documentation for simd_shuffle4() but
I believe packed_simd implements exactly the same interface.

Plus, implementing this change has fixed font-kit glyph output on
an Apple M1 mac (64-bit arm).

Closes #450

### DIFF
--- a/simd/src/arm/mod.rs
+++ b/simd/src/arm/mod.rs
@@ -129,7 +129,7 @@ impl F32x2 {
 
     #[inline]
     pub fn concat_xy_xy(self, other: F32x2) -> F32x4 {
-        unsafe { F32x4(simd_shuffle4(self.0, other.0, [0, 1, 0, 1])) }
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Looks reasonable; thanks!

PR #466, opened 2021-06-05
Update canvas_nanovg example

laptou:
The `canvas_nanovg` example would not run on my Linux machine due to the bug in `winit` 0.19 (rust-windowing/winit#1773)

In this PR, I updated `winit` and `surfman` and made the minimum necessary changes so that the example would compile and run.


jdm:
@bors-servo r+
Thanks!

PR #487, opened 2022-01-11
Make the demo compile with NDK

floppyhammer:
Solving #472. Upgrade `usvg` to `0.20.0`, which removes `harfbuzz_rs` dependency.

PR #493, opened 2022-02-05
Upgrade usvg from 0.9 -> 0.20

kognise:
Upgrades `usvg` to 0.20. Had to make some minor code changes, the main differences are:

- `Tree` constructors now take an `OptionsRef` instead of an `Options`, so we need a bunch of `.to_ref()` calls
- `Tree::from_file` was removed, I replaced it with `Tree::from_data` and `std::fs::read`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "ab_glyph_rasterizer"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a13739d7177fbd22bb0ed28badfff9f372f8bef46c863db4e1c6248f6b223b6e"
+
+[[package]]
 name = "addr2line"
 version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


kognise:
Just realized #487 does mostly the same thing, I didn't notice since it has a different name. I believe my changes fixed a couple more things than it, but I'm not 100% sure.

floppyhammer:
@kognise The focus of https://github.com/servo/pathfinder/pull/487 is to fix the Android demo. So, I don't mind rebasing that on your commit if yours does more fixes. In addition, I've got a little question. Is it necessary to commit `Cargo.lock`?

kognise:
A rebase might be a good idea, up to you. As for `Cargo.lock`, everything breaks if we *don't* commit it so I think it's fine. Has the potential to cause merge conflicts though, so I'm not 100% sure.

floppyhammer:
@kognise Do you have any time to work on this?

kognise:
Sorry, I'll try to handle it by Monday at the latest.

kognise:
Trying to get these PRs working... I'm stuck on an M1 MacBook right now, which is normally a blessing!

I'm having trouble with the `winit` versions depended on— neither `0.21` nor `0.19` will build on aarch64 as the `objc` crate introduced some breaking arch-specific type changes (an `i8` to a `bool`, but only on M1). This has been fixed on the latest version of winit (`0.24.0`) but I'm struggling to navigate the dependency nightmare.

Even after upgrading Winit in all of Pathfinder, Surfman still depends on the older version. What's the best way to handle this? Should I make a PR to Surfman and wait for that to merge?

floppyhammer:
@kognise It seems the `mismatched types` issue already exists on the current master. It might need another PR.

PR #495, opened 2022-02-08
Fix links in docs, fix/ignore some doctests

5225225:


5225225:
Does this project have CI? If it does, then might be worth adding a `cargo doc` run to find warnings like this. (I'd need to look up the exact command)

jdm:
Oh, shoot, I forgot that we disabled travisci and haven't migrated to GitHub actions yet. I knew I was putting that off for a good reason.

PR #507, opened 2022-03-29
Don't use hardcoded bin workgroup size

floppyhammer:


jdm:
@bors-servo r+
Thanks!

PR #518, opened 2022-07-21
The scope of the unsafe block can be appropriately reduced

cactter:


s3bk:
`unsafe` blocks encapsulate the code that needs to be correct to ensure soundness.
The value and type of `vector` are critical here, so reducing the block is (in my opinion) not correct.

PR #552, opened 2024-02-12
Update the PDF dependency

s3bk:


s3bk:
It builds now, but crashes on wayland. (That also happened before.)

mrobinson (approved):

PR #555, opened 2024-02-15
Fix #553

### LINKED ISSUE #553: Panicking when font is not found or can't be loaded
I'm using the Pathfinder C-binding to render text in my application, and I noticed that the program crashed irrecoverably when the selected font wasn't found or couldn't be loaded. I initially thought this was just a minor problem in the C-binding code, but when I investigated it further, it seems to be a deeper design issue.

https://github.com/servo/pathfinder/blob/e4fcda0d5259d0acf902aee6de7d2501f2bd6629/canvas/src/text.rs#L232-L236

I'm not a Rust expert so I don't know the reason for these panics, but I think Pathfinder should propagate these error conditions so that they can be handled further up by the caller, including from C.

s3bk:
This replaces two unwraps with Results and makes `PFCanvasSetFontByPostScriptName` return an i32 to indicate success or failure.
Intended to fix #553

### DIFF
--- a/c/src/lib.rs
+++ b/c/src/lib.rs
@@ -377,8 +377,11 @@ pub unsafe extern "C" fn PFCanvasSetLineDashOffset(canvas: PFCanvasRef, new_offs
 #[no_mangle]
 pub unsafe extern "C" fn PFCanvasSetFontByPostScriptName(canvas: PFCanvasRef,
                                                          postscript_name: *const c_char,
-                                                         postscript_name_len: usize) {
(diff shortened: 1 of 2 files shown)


jayvdb (approved):
lgtm, just noting that these are breaking changes, but better done sooner than later

s3bk:
Yea. This isn't really possible without breaking existing code.
I modified it a bit so the underlying errors are available and the  `unwrap()` works.

s3bk:
@jayvdb can you review it again?

jayvdb (approved):

jdm (approved):

PR #548, opened 2024-02-06
remove stdsimd feature flag

jonaspleyer:
# Problem

I was recently compiling the [plotters](https://github.com/plotters-rs/) crate with the nigthly compiler and got the following build error:

```
error[E0635]: unknown feature `stdsimd`
  --> /home/REDACTED/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pathfinder_simd-0.5.2/src/lib.rs:12:49
   |
12 | #![cfg_attr(pf_rustc_nightly, feature(simd_ffi, stdsimd))]
   |                                                 ^^^^^^^

For more information about this error, try `rustc --explain E0635`.
```
My nightly compiler is
```
# rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/REDACTEDrustup

installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu
nightly-2022-12-18-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu (default)
1.58-x86_64-unknown-linux-gnu
1.75-x86_64-unknown-linux-gnu
1.62.0-x86_64-unknown-linux-gnu

active toolchain
----------------

nightly-x86_64-unknown-linux-gnu (default)
rustc 1.78.0-nightly (f067fd608 2024-02-05)
```

## Solution

It seems that the feature flag `stdsimd` is not present in the nightly compiler anymore.
This patch very small patch is able to fix my problem and the package still compiles
```
diff --git a/simd/src/lib.rs b/simd/src/lib.rs
index cbffc2a4..89a087bd 100644
--- a/simd/src/lib.rs
+++ b/simd/src/lib.rs
@@ -9,7 +9,7 @@
 // except according to those terms.
 
 #![cfg_attr(pf_rustc_nightly, feature(link_llvm_intrinsics, platform_intrinsics))]
-#![cfg_attr(pf_rustc_nightly, feature(simd_ffi, stdsimd))]
+#![cfg_attr(pf_rustc_nightly, feature(simd_ffi))]
 
 //! A minimal SIMD abstraction, usable outside of Pathfinder.
```
with this stable compiler:
```
# rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/REDACTED.rustup

installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu (default)
nightly-2022-12-18-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu
[...]

PR #557, opened 2024-02-25
bump simd to 0.5.3 in order push fixes for stdsimd feature upstream

jonaspleyer:
This is a follow-up of #548 to push simd upstream.

### DIFF
--- a/simd/Cargo.toml
+++ b/simd/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pathfinder_simd"
-version = "0.5.2"
+version = "0.5.3"
 edition = "2018"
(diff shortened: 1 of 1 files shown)


jonaspleyer:
@s3bk How do we proceed from here? Are there any more steps that need to be taken? Please let me know. I am looking forward for this fix in my other crates.

jonaspleyer:
Tbh I did not know that I can do that. I will be looking out for that in the future though. Thanks again!

Maurdekye:
when will this be published to crates.io?

atouchet:
@Maurdekye it is available now: https://crates.io/crates/pathfinder_simd/0.5.3

PR #575, opened 2025-02-13
Apply @/theoparis' fixes for arm simd

itsjunetime:
This fixes https://github.com/servo/pathfinder/issues/571 by applying the changes found in https://github.com/theoparis/pathfinder. I'm not certain why @theoparis did not upstream these themself, but I would like this fix upstreamed, so I'm filing this.

I haven't tested this with versions of nightly before the rust commit that broke this code, so it might break those, but I expect there are more people are using the most recent version of nightly than people using versions that still work with pathfinder_simd.


theoparis:
I didn't upstream this because I didn't realize these changes would work on stable rust with  `#[repr(simd)]` feature-gated.

itsjunetime:
It does require nightly, but it seems that pathfinder has their own `pf_rustc_nightly` cfg to enable compilation that is conditional upon running it under nightly. I didn't realize that at first either, it took me a second of poking to see.

strottos:
Hit this on my error on my ARM based Windows laptop from a different crate and this fix fixes things for me. Would approve of this being merged (or an indication of what needs to be done to be satisfied that this can be merged perhaps?).

jdm (approved):

ahcm:
As this breaks plotters via the font-kit dependancy, a release would be very welcome!

PR #576, opened 2025-03-14
Fix Linux Build

cmrschwarz:
This PR allows the pathfinder workspace to compile and pass all tests on Linux again.

I find it very helpful to be able to run `cargo test --all` in a workspace just to check if I broke anything.

There were three main issues preventing this:

1. The `canvas_metal_minimal` example expects to link against metal specific APIs:
    I disabled this example on non-apple platforms in a way that's similar to how other projects like `rust-skia` do this.

2.  The android demo had some minor inconsistencies (seems to not have been compiled in a while ... ;)),
     so I fixed those.

3. The Magic Leap demo is pretty fundamentally broken. It uses nonexistant APIs and expects to call into many now private internals.
     I decided to comment out it's member entry in the root workspace.
    Since the hardware it was written for seems to be [no longer usable, even for preexisting owners](https://en.wikipedia.org/wiki/Magic_Leap), I was hoping this would be an acceptable tradeoff. It certainly does not seem to make things worse.
    
With these 3 small changes you can now `cargo test --all` and `cargo build --all` the project on Linux again.


jdm (approved):
These are all reasonable choices. Thanks!

cmrschwarz:
Awesome, thanks <3.

I know this project has went through a bit of a rough patch, but I still find it very useful even in its current state.

Is there any chance we could get the  `pathfinder_webgl`  and maybe even `pathfinder` (https://github.com/pontuslaestadius/pathtracer/issues/41) crates released on crates io @jdm? 

I would be happy to help out with any of the necessary work.

PR #581, opened 2025-07-03
simd: Round with round_ties_even

### LINKED ISSUE #579: rounding in scalar implementation of to_i32x4 differs from SSE implementation
The SSE implementation uses `cvtps2dq` which uses the round to nearest mode by default. The scalar implementation calls `round`, which always rounds away from zero.

https://github.com/servo/pathfinder/blob/974bc994159beee2f8bd959ddc9771519af3f729/simd/src/scalar/mod.rs#L320-L327

The difference can be observed with this program:

```rust
fn main() {
    let a = pathfinder_simd::x86::F32x4::new(20.5, -8.96875, 69.1875, 55.09375);
    let b = pathfinder_simd::scalar::F32x4::new(20.5, -8.96875, 69.1875, 55.09375);

    println!("simd: {:?}", a.to_i32x4());
    println!("sclr: {:?}", b.to_i32x4());
}
```

On an x86_64 host, when run with `cargo r` the first point (20.5) is rounded to 20 in the SIMD implementation, but 21 in the scalar implementation.

This was observed in practice as differing output in x86_64 machines that use the simd implementation be default and aarch64 machines that currently use the scalar implementation by default.

I _think_ switching from [round](https://doc.rust-lang.org/std/primitive.f32.html#method.round) to [round_ties_even](https://doc.rust-lang.org/std/primitive.f32.html#method.round_ties_even) in the scalar implementation would address the discrepancy. I can make this change if it seems reasonable.

wezm:
This makes the rounding behaviour match the x86 SIMD implementation.

Fixes #579

### DIFF
--- a/simd/src/scalar/mod.rs
+++ b/simd/src/scalar/mod.rs
@@ -119,13 +119,13 @@ impl F32x2 {
     /// Converts these packed floats to integers via rounding.
     #[inline]
     pub fn to_i32x2(self) -> I32x2 {
-        I32x2([self[0].round() as i32, self[1].round() as i32])
+        I32x2([self[0].round_ties_even() as i32, self[1].round_ties_even() as i32])
     }
 
     /// Converts these packed floats to integers via rounding.
     #[inline]
     pub fn to_i32x4(self) -> I32x4 {
-        I32x4([self[0].round() as i32, self[1].round() as i32, 0, 0])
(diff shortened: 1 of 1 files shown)


wezm:
Any chance a new release could be made with this change included? I'd like to make a new [Allsorts](https://github.com/yeslogic/allsorts) release, but need an updated published version to depend on.

PR #583, opened 2026-03-31
Fix ARM nightly compilation

itsjunetime:
Currently, this crate fails to compile with the latest nightly due to some changes to simd intrinsic names (see https://github.com/rust-lang/rust/commit/986a28064468549727952e290d4eb062b17c68d2). So this makes them work with the latest compiler.

This does make this crate fail to compile with an older nightly, but keeping with latest nightly seems to be pretty par-for-the-course for this crate, so I'm not worried about that.

I also changed some test constants to match what llvm now outputs. The new constants are pretty close to what were being output before, so I'm not worried about this either; these values were already considered 'approximate' (judging by the method name), so I think it's fine.

### DIFF
--- a/simd/src/arm/mod.rs
+++ b/simd/src/arm/mod.rs
@@ -79,12 +79,12 @@ impl F32x2 {
 
     #[inline]
     pub fn min(self, other: F32x2) -> F32x2 {
-        unsafe { F32x2(simd_fmin(self.0, other.0)) }
+        unsafe { F32x2(simd_minimum_number_nsz(self.0, other.0)) }
     }
 
     #[inline]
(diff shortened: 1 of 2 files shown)


itsjunetime:
Hmm, actually, this is interesting - it seems that the `F32x4::approx_recip` fn returns different results based on if it's nightly or not. The llvm intrinsic version results in answers that are slightly off, but the manual division (non-intrinsic) version that we use on stable results in more accurate answers. I've added a range comparison instead, to make sure they're within `0.002` of the correct answer each time.

sagudev (approved):
Thank you!

PR #589, opened 2026-05-02
Fix I32x4 min/max working incorrectly for large values

Shnatsel:
The ARM I32x4::{min,max} implementations must compare integer lanes directly. Converting through f32 loses precision above 2^24, so values such as 16_777_217 round to neighboring integers and can produce a numerically wrong vector.

This PR fixes the issue and adds a regression test. The issue was spotted by GPT-5.5 xhigh.

### DIFF
--- a/simd/src/arm/mod.rs
+++ b/simd/src/arm/mod.rs
@@ -604,12 +604,12 @@ impl I32x4 {
 
     #[inline]
     pub fn max(self, other: I32x4) -> I32x4 {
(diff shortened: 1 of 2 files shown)


sagudev:
this looks related to https://github.com/servo/pathfinder/pull/583 and I suspect we can also revert test changes.

> This PR fixes the issue and adds a regression test. The issue was spotted by GPT-5.5 xhigh.

But the PR/code was not done by AI, right?

Shnatsel:
The regression test/PoC is AI-written. I always have models write a PoC that fails miri before looking into the issue.

The production code change is human.

Feel free to drop the test if you have a policy against including AI contributions.

